PR #17, opened 2019-08-11
WIP: making it run from Dockerfile/Image

mfbieber:
- RenderController: by default port 8080 is used
- ThreeDRenderController: get 3d.html also if not running from IntelliJ
- Dockerfile: use not slim base image, because libfontmanager needs libfreetype (java.lang.UnsatisfiedLinkError: /usr/local/openjdk-11/lib/libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory)
- pom: added apacke commons-io to use IOUtils

### DIFF
--- a/src/main/java/de/bonndan/nivio/output/threeD/ThreeDRenderController.java
+++ b/src/main/java/de/bonndan/nivio/output/threeD/ThreeDRenderController.java
@@ -1,8 +1,12 @@
 package de.bonndan.nivio.output.threeD;
 
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
 import de.bonndan.nivio.api.NotFoundException;
 import de.bonndan.nivio.landscape.Landscape;
 import de.bonndan.nivio.landscape.LandscapeRepository;
+import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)


mfbieber:
I am building the docker image locally and running it with: `docker run -e SEED=/tmp/base-configuration.yml --mount type=bind,source="$(pwd)"/config,target=/tmp -p 8089:8080 nivio:local1`

src/main/java/de/bonndan/nivio/output/threeD/ThreeDRenderController.java
```diff
@@ -29,11 +29,11 @@
     @Autowired
     public ThreeDRenderController(LandscapeRepository landscapeRepository) throws IOException {
         this.landscapeRepository = landscapeRepository;
-        content = new String(Files.readAllBytes(new File("src/main/resources/static/html/3d.html").toPath()));
+        content = IOUtils.toString(getClass().getResourceAsStream("/static/html/3d.html"), StandardCharsets.UTF_8);
     }
 
     @RequestMapping(method = RequestMethod.GET, path = "/render/{landscape}/threeD.html")
-    public ResponseEntity<String> three3Resource(@PathVariable(name = "landscape") final String landscapeIdentifier) throws IOException {
+    public ResponseEntity<String> three3Resource(@PathVariable(name = "landscape") final String landscapeIdentifier) {
```
mfbieber: That exception was not thrown according to IntelliJ - am I overlooking something?

src/main/java/de/bonndan/nivio/output/threeD/ThreeDRenderController.java
```diff
@@ -29,7 +29,7 @@
     @Autowired
     public ThreeDRenderController(LandscapeRepository landscapeRepository) throws IOException {
         this.landscapeRepository = landscapeRepository;
-        content = new String(Files.readAllBytes(new File("src/main/resources/static/html/3d.html").toPath()));
+        content = IOUtils.toString(getClass().getResourceAsStream("/static/html/3d.html"), StandardCharsets.UTF_8);
```
bonndan: :+1: for using resourceAsStream

However, I'd prefer not to introduce a new dependency for resource loading. There might be something from Spring or Java ready to use.

PR #20, opened 2019-08-12
WIP: setting colors from the ouside

mfbieber:
added configMap and started to set color of services from yml configuration files


bonndan:
There is a solution in the master branch now. Please try.

Issue #27, opened 2019-09-26
Configurable greediness/gracefulness


bonndan:
Currently unknown Relation (dataflow, Provider) targets are ignored. The Landscape config should allow to add unknown Elements in the fly instead of ignoring them.

Affects:
* relations
* Machines?

- [ ] change default behaviour to be greedy, i.e. create everything instead of rejecting items

bonndan:
Not changing default behavior.

PR #48, opened 2019-11-19
[#27] unknown providers and relation targets are added on the fly

bonndan:


src/main/java/de/bonndan/nivio/input/InstantItemResolver.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public InstantItemResolver(ProcessLog log) {
+        this.log = log;
+    }
+
+    public void processTargets(LandscapeDescription landscape) {
+        List<ItemDescription> all = landscape.getItemDescriptions();
+
+        if (!landscape.getConfig().isGreedy())
+            return;
```
basseur: This should be the first thing to be called before instantiating `List<ItemDescription> all`.

Also, I personally would never omit curly braces. For guard clauses like this, it _may_ be okay, if you move the return onto the same line.

bonndan: :+1:

src/main/java/de/bonndan/nivio/input/Indexer.java
```diff
@@ -53,6 +53,7 @@ public ProcessLog reIndex(final LandscapeDescription input) {
             Map<ItemDescription, List<String>> templatesAndTargets = new HashMap<>();
             new SourceReferencesResolver(formatFactory, logger).resolve(input, templatesAndTargets);
             new TemplateResolver().processTemplates(input, templatesAndTargets);
+            new InstantItemResolver(logger).processTargets(input);
```
bonndan: @basseur @Matthimatiker @sl0815 

Any suitable patterns here?

basseur: What do you mean? 😬

bonndan: new X, new Y, new Z ... and probably the group creation below might be handled by a class, too.

Issue #51, opened 2019-11-22
Blacklist stacks in (rancher) landscape


bonndan:
The architect may not want to see "infra" stacks containing rancher-internal stuff.

bonndan:
stacks need to be filtered based on their "externalId" value.

     "externalId": "catalog://library:infra*network-services:29",

PR #64, opened 2019-11-28
[#51] blacklisting of groups

bonndan:


docs/source/install.rst
```diff
@@ -75,6 +75,7 @@ For jgraphx output, some force directed graph params can be set. More configurat
 
     # landscape configuration
     config:
+      groupBlacklist: [".*infra.*", "hidden"]
```
sl0815: 👍

src/main/java/de/bonndan/nivio/input/GroupResolver.java
```diff
@@ -19,24 +28,52 @@ public GroupResolver(ProcessLog log) {
...
 
-            String group = itemDescription.getGroup();
-            if (StringUtils.isEmpty(itemDescription.getGroup())) {
+            String group = item.getGroup();
+            if (StringUtils.isEmpty(item.getGroup())) {
                 group = Group.COMMON;
             }
 
-            landscape.getGroups().computeIfAbsent(group, s -> getGroup(s, null));
+            if (!isBlacklisted(group, specs))
```
sl0815: On lines 37 - 42 you use curly braces for the if/else statement, but here not? Personally, I wouldn't omit them. But either way you should act consistenlty.

bonndan: has been fixed, thanks

PR #77, opened 2020-01-08
76 diet

bonndan:
This reduces the uberjar size by about 30MB (to about 40MB).


src/main/java/de/bonndan/nivio/output/Color.java
```diff
@@ -38,10 +37,10 @@ public static String nameToRGB(String name, String defaultColor) {
 
     public static String lighten(String color) {
         try {
-            java.awt.Color col = WebColors.getRGBColor("#" + color).brighter();
-            return Integer.toHexString(col.getRGB());
+            java.awt.Color col = java.awt.Color.decode(color.startsWith("#") ? color : "#" + color);
```
basseur: Maybe you could choose a different name for your `Color` class... I don't know about best practice here, but I personally don't like using fully qualified names (`java.awt.Color`), if it can be avoided.

src/main/java/de/bonndan/nivio/output/docs/DocsController.java
```diff
@@ -35,35 +30,6 @@ public DocsController(LandscapeRepository landscapeRepository, IconService iconS
...
     }
 
-    @RequestMapping(method = RequestMethod.GET, path = "/{landscape}")
-    public ResponseEntity<String> docResource(@PathVariable(name = "landscape") final String landscapeIdentifier) {
-
-        LandscapeImpl landscape = landscapeRepository.findDistinctByIdentifier(landscapeIdentifier).orElseThrow(
-                () -> new NotFoundException("Landscape " + landscapeIdentifier + " not found")
-        );
-
-        Map<String, Object> attributes = new HashMap<String, Object>();
```
basseur: You can also remove the imports of `Map` and `HashMap`.

src/main/java/de/bonndan/nivio/output/Color.java
bonndan: Afaik there is no way around fully qualified names when the class has the same name.

basseur: That iss correct. That's why I was asking if you'd change the name. I guess this could be a matter of taste.

PR #100, opened 2020-02-03
Compile time file imports

Matthimatiker:
Allows to import files (e.g. templates) in components at compile time.


bonndan:
A theoretical question, since JQuery is used in another place: Could it be loaded at compile time instead of runtime in Man.js?

bonndan:
@Matthimatiker @MarvinSchoening I think this has become obsolete with Marvin's refactorings. Can we close it?

Matthimatiker:
Perhaps we can rescue the module documentation.

PR #115, opened 2020-03-09
fix tests for windows

### LINKED ISSUE #112: Fix tests on windows
Some tests fail due to hardcoded forward slashes.

MarvinSchoening:
Closes #112

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ _static
 _templates
 /.env
 *.iml
+src/test/resources/example/*_debug.png
--- a/src/test/java/de/bonndan/nivio/input/LandscapeDescriptionFactoryTest.java
+++ b/src/test/java/de/bonndan/nivio/input/LandscapeDescriptionFactoryTest.java
@@ -11,6 +11,7 @@ import org.junit.jupiter.api.Test;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


src/test/java/de/bonndan/nivio/input/SeedTest.java
```diff
@@ -47,10 +48,18 @@ public void fails() throws MalformedURLException {
 
     @Test
     public void windowsFileLocation() throws MalformedURLException {
-        Seed seed = new Seed("c:\\a\\b\\c.yml");
+        final String separator = FileSystems.getDefault().getSeparator();
+        final String seedLocation = "c:" + separator + "a" + separator + "b" + separator+ "c.yml";
+        /*
+        TODO: Should it really be file://c/a/b/c.yml? We remove the : from C: in our Seed class but if i want to open the URL without : it wont work in my browser, but dont know whats happening internally with the URL
```
bonndan: We could try to hide internals from the user, i.e. allowing anything that could be a local url, like "c:\foo\bar.yml", "./foo.yml". Given the scope of this ticket, I'd say we'd open another one to improve user experience.

src/test/java/de/bonndan/nivio/input/LandscapeDescriptionFactoryTest.java
```diff
@@ -23,17 +24,26 @@
 
 
 class LandscapeDescriptionFactoryTest {
+    final String SEPARATOR = FileSystems.getDefault().getSeparator();
```
bonndan: I'd prefer to define only reusable constants here.

.gitignore
```diff
@@ -5,3 +5,4 @@ _static
 _templates
 /.env
 *.iml
+src/test/resources/example/*_debug.png
```
bonndan: Sure? It might be interesting to see how code changed affect renderings, but I also see that invisible changes are annoying.

bonndan: agreed

Issue #114, opened 2020-03-05
Remodel landscape components


bonndan:
- [x] store anything in items that can be expressed as key-value
- [x] discard "known" fields that only make sense in a few cases (e.g. host type)
- [x] keep links in items and add a meaning (like in source references pointing at k8s) #97

PR #121, opened 2020-04-10
114 remodel

bonndan:
This would be nivio 0.3.0 and implements #114


src/main/java/de/bonndan/nivio/assessment/Assessable.java
```diff
@@ -0,0 +1,66 @@
...
+
+    Set<StatusValue> getStatusValues();
+
+    default void setStatusValue(StatusValue statusValue) {
+
+        if (statusValue == null) {
+            throw new IllegalArgumentException("Status value is null");
+        }
+        if (StringUtils.isEmpty(statusValue.getField())) {
+            throw new IllegalArgumentException("Status value has no field");
```
Matthimatiker, line 53: I think this check could also be moved to the ``StatusValue`` constructor. That would ensure that instances of that class are always valid.

src/main/java/de/bonndan/nivio/assessment/Assessable.java
```diff
@@ -0,0 +1,66 @@
...
+
+        return new StatusValue(SUMMARY_LABEL + "." + getIdentifier(), summary.get().getStatus(), summary.get().getField());
+    }
+
+    Set<StatusValue> getStatusValues();
+
+    default void setStatusValue(StatusValue statusValue) {
+
+        if (statusValue == null) {
+            throw new IllegalArgumentException("Status value is null");
```
Matthimatiker, line 50: :+1: 

I often use ``Objects.requireNonNull()`` as I am lazy. But this check provides the slightly better message.

src/main/java/de/bonndan/nivio/assessment/Assessment.java
```diff
@@ -0,0 +1,25 @@
+package de.bonndan.nivio.assessment;
+
+import de.bonndan.nivio.assessment.kpi.KPI;
+
+import java.util.List;
+import java.util.Map;
+
+public class Assessment {
```
Matthimatiker, line 8: A short description of the class would be helpful.

src/main/java/de/bonndan/nivio/assessment/Assessable.java
```diff
@@ -0,0 +1,66 @@
...
+                summary.set(value);
+            }
+        });
+
+        return new StatusValue(SUMMARY_LABEL + "." + getIdentifier(), summary.get().getStatus(), summary.get().getField());
+    }
+
+    Set<StatusValue> getStatusValues();
+
+    default void setStatusValue(StatusValue statusValue) {
```
Matthimatiker, line 47: This is more an "add" than a "set".

src/main/java/de/bonndan/nivio/assessment/Assessable.java
```diff
@@ -0,0 +1,66 @@
...
+
+        if (statusValue == null) {
+            throw new IllegalArgumentException("Status value is null");
+        }
+        if (StringUtils.isEmpty(statusValue.getField())) {
+            throw new IllegalArgumentException("Status value has no field");
+        }
+
+        getStatusValues().add(statusValue);
```
Matthimatiker, line 56: This line assumes that the implementation provides a mutable ``Set``.

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.lang.NonNull;
+import org.springframework.lang.Nullable;
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * A configured key performance indicator related to a landscape item label.
```
Matthimatiker, line 19: :+1:

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+public class CustomKPI extends KPI {
```
Matthimatiker, line 21: Renaming ``KPI`` to ``KeyPerformanceIndicator`` would make 100% clear what is meant.

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
```diff
@@ -0,0 +1,138 @@
...
+
+    public static final String SEPARATOR = ";";
+    private final Map<Status, Range<Double>> ranges;
```
Matthimatiker, line 24: I like the ``Status`` as key.

Nevertheless, the map structure would profit from some documentation.

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private Optional<Status> getStatusByRange(String value) {
+        if (value == null) {
+            return Optional.empty();
+        }
+
+        double d;
+        try {
+            d = Double.parseDouble(value);
+        } catch (NumberFormatException ignored) {
```
Matthimatiker, line 112: Is there a valid case where it is ok to ignore this exception? Doesn't it hide more errors?

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
```diff
@@ -0,0 +1,138 @@
...
+            return Optional.empty();
+        }
+        return ranges.entrySet().stream()
+                .filter(entry -> entry.getValue().contains(d))
+                .findFirst()
+                .map(Map.Entry::getKey);
+    }
+
+    private void addSpecs(@Nullable Map<Status, String> statusPatterns) {
```
Matthimatiker, line 121: :+1:  for ``@Nullable``

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
```diff
@@ -0,0 +1,138 @@
...
+        if (statusPatterns == null)
+            return;
+
+        statusPatterns.forEach((status, strings) -> {
+            List<Function<String, Boolean>> specs = Arrays.stream(strings.split(SEPARATOR)).map(s -> {
+                try {
+                    Pattern p = Pattern.compile(s, Pattern.CASE_INSENSITIVE);
+                    return (Function<String, Boolean>) s1 -> p.matcher(s1).matches();
+                } catch (Exception e) {
```
Matthimatiker, line 130: Can we catch a more specific exception here? Can we check before if the pattern can be compiled?

Is the fallback really the desired behavior? What if I supplied an invalid regex (because of a missing brace etc.), but never intended to apply the ``contains`` behavior?

src/main/java/de/bonndan/nivio/input/LabelToFieldProcessor.java
```diff
@@ -79,9 +81,15 @@ private void setValue(ItemDescription item, String name, String value) {
...
-            myAccessor.setPropertyValue(name, value.trim());
+            if (name.startsWith(Linked.LINK_LABEL_PREFIX)) {
+                item.setLink(name.replace(Linked.LINK_LABEL_PREFIX, ""), new URL(value));
+            } else {
+                myAccessor.setPropertyValue(name, value.trim());
+            }
         } catch (NotWritablePropertyException e) {
             logger.warn("Failed to write field '" + name + "' via label");
+        } catch (MalformedURLException e) {
+            logger.warn("Failed to add link '" + name + "' via label because of malformed URL " + value);
```
Matthimatiker, line 92: You can pass the exception as last parameter to ensure that it is not completely lost.

src/main/java/de/bonndan/nivio/input/dto/ItemDescription.java
```diff
@@ -397,15 +230,59 @@ public String toString() {
...
+        if (value instanceof String) {
+            labels.put(key.toLowerCase(), (String) value);
+            return;
+        }
+        if (value instanceof String[]) {
+            Arrays.stream(((String[]) value)).forEach(s -> setPrefixed(key, s));
+            return;
+        }
+
+        if (value instanceof List) {
```
Matthimatiker, line 273: Is it possible to avoid the checks and casts by overloading the method?

src/main/java/de/bonndan/nivio/model/Labeled.java
```diff
@@ -1,13 +1,149 @@
...
+
+    /**
+     * Returns all labels with the given prefix.
+     *
+     */
+    default Map<String, String> getLabels(Label prefix) {
+        return getLabels(prefix.name().toLowerCase());
+    }
+
+    default void setLabels(String prefix, String[] keys, String[] values) {
```
Matthimatiker, line 38: Why are ``keys`` and ``values`` not passed as ``Map``? Because it looks as if that is expected.

src/main/java/de/bonndan/nivio/model/RelationType.java
```diff
@@ -20,4 +24,13 @@ public static RelationType from(String relationType) {
         return DATAFLOW;
     }
 
+    /**
+     * Filters a list of relations for equal type.
+     *
+     * @param all all relations
+     * @return filtered subset
+     */
+    public <T> List<RelationItem> filter(Collection<? extends RelationItem> all) {
```
Matthimatiker, line 33: Where is ``T`` used?

src/test/java/de/bonndan/nivio/assessment/AssessableTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class AssessableTest {
```
Matthimatiker, line 13: :+1:

src/test/java/de/bonndan/nivio/assessment/StatusValueTest.java
```diff
@@ -0,0 +1,28 @@
...
+class StatusValueTest {
+
+    @Test
+    void highestOf() {
+        List<StatusValue> statusValueSet = new ArrayList<>();
+        statusValueSet.add(new StatusValue("foo", Status.GREEN));
+        statusValueSet.add(new StatusValue("bar", Status.ORANGE));
+        statusValueSet.add(new StatusValue("baz", Status.ORANGE));
+
+        List<StatusValue> highest = StatusValue.highestOf(statusValueSet);
```
Matthimatiker, line 19: Did not notice that before, but I would expect ``highestOf()`` to return a single item (or an ``Optional`` of an item).

src/test/resources/example/dedica.yml
```diff
@@ -0,0 +1,54 @@
+identifier: dedica.team
```
Matthimatiker, line 1: Cool :-)

src/main/java/de/bonndan/nivio/api/ApiController.java
```diff
@@ -60,19 +61,22 @@ public ApiController(LandscapeRepository landscapeRepository, ItemDescriptionFor
...
-    @RequestMapping(path = "/landscape/{identifier}")
+    @RequestMapping(path = "/landscape/{identifier}", produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<LandscapeDTO> landscape(@PathVariable String identifier) {
         LandscapeImpl landscape = landscapeRepository.findDistinctByIdentifier(identifier).orElse(null);
         if (landscape == null)
             return ResponseEntity.notFound().build();
-
-        return new ResponseEntity<>(createDTO().apply(landscape), HttpStatus.OK);
+        Assessment assessment = new Assessment(landscape.getConfig().getKPIs());
+        assessment.assess(landscape);
```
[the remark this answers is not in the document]
bonndan: I had in mind to decouple assessment (volatile state) from the landscape (much more static). This could be also worth an extra API endpoint (?)

But then again there is the Assessable interface which encourages to set state on landscape components. Any good idea? @Matthimatiker

src/main/java/de/bonndan/nivio/api/LandscapeDTOFactory.java
```diff
@@ -21,57 +24,52 @@
...
-        l.description = landscape.getDescription();
-
-        if (landscape instanceof LandscapeImpl) {
-            l.stats = getLandscapeStats((LandscapeImpl) landscape);
-        }
-
-        return l;
+    static {
+        mapper = new ObjectMapper();
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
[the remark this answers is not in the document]
bonndan: This code was obsolete, sorry.

src/main/java/de/bonndan/nivio/api/LandscapeDTOFactory.java
```diff
@@ -21,57 +24,52 @@
...
 
-        LandscapeStatistics stats = new LandscapeStatistics();
-        stats.items = impl.getItems().all().size();
-        stats.groups = impl.getGroups().size();
-
-        List<StatusItem> collect = impl.getItems().stream()
-                .map(item -> StatusItem.highestOf(item.getStatuses())).flatMap(Collection::stream)
-                .collect(Collectors.toList());
+        LandscapeDTO landscapeDTO = new LandscapeDTO();
+        if (landscape == null)
```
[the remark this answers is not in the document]
bonndan: org.springframework.lang also provides these annotations. Is findbugs more suitable?

src/main/java/de/bonndan/nivio/api/LandscapeDTOFactory.java
```diff
@@ -21,57 +24,52 @@
...
             if (messages.size() > 0) {
-                stats.lastUpdate = messages.get(messages.size() - 1).getDate();
+                landscapeDTO.lastUpdate = messages.get(messages.size() - 1).getDate();
             }
         }
-        return stats;
+        return landscapeDTO;
+    }
+
+    private static Map<String, Group> getGroups(LandscapeImpl landscape) {
```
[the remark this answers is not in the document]
bonndan: done

src/main/java/de/bonndan/nivio/model/Labeled.java
bonndan: Method has been removed.

src/main/java/de/bonndan/nivio/api/dto/LandscapeDTO.java
```diff
@@ -1,62 +1,49 @@
...
     public String name;
     public String contact;
     public String description;
 
     public String source;
-    public Map<String, GroupItem> groups;
-    public LandscapeStatistics stats;
+    public Map<String, Group> groups;
+    public Set<Item> items;
+    public Date lastUpdate;
```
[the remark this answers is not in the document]
bonndan: Done via merge from develop branch

src/main/java/de/bonndan/nivio/api/LandscapeDTOFactory.java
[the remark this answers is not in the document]
bonndan: Is it all JSR-305?

src/main/java/de/bonndan/nivio/assessment/Assessable.java
bonndan: done

src/main/java/de/bonndan/nivio/assessment/Assessable.java
bonndan: The concept behind it is a distinct Set, determined by StatusValue.field

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
bonndan: done

src/main/java/de/bonndan/nivio/input/dto/ItemDescription.java
bonndan: This JsonAnySetter is a last resort anways. See also https://stackoverflow.com/questions/22501844/multiple-any-setters-using-jackson

But yes, that would be better

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
bonndan: We do not know the value to assess, which can be anything and can be supplied even by a third party system. So first we try parsing numbers to Double, and it that fails we assume it is a string and is compared against the regex matchers.

One improvement could be throwing an exception if only numeric ranges are defined for the KPI, but then we'd have to handle the exception, and a wrong value from a third party system is not necessarily exceptional.

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
bonndan: But we need to save precious bytes!

src/main/java/de/bonndan/nivio/input/LabelToFieldProcessor.java
bonndan: This is nivio's special Logger decorator, meaning everything logged here lands in the GUI

src/main/java/de/bonndan/nivio/model/RelationType.java
bonndan: removed

src/test/java/de/bonndan/nivio/assessment/StatusValueTest.java
bonndan: No, for instance a landscape item could have Status.RED concerning security AND health, and neither of them is higher than the other. The method is unused yet, however.

src/main/java/de/bonndan/nivio/assessment/kpi/CustomKPI.java
bonndan: What would you suggest? Simply warn the user that the regex is wrong?

src/main/java/de/bonndan/nivio/assessment/Assessment.java
bonndan: done

src/main/java/de/bonndan/nivio/api/LandscapeDTOFactory.java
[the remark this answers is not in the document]
Matthimatiker: > Is findbugs more suitable?

I never used the annotation from Spring, so I cannot give a comparison.
I like that the annotations from findbugs are purely informational.

In contrast, some similar annotations from Lombok add additional behavior.

src/main/java/de/bonndan/nivio/api/ApiController.java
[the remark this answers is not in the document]
Matthimatiker: I would need more information to make a qualified decision here.

Separating the static landscape from the state sound reasonable.

Regarding `Assessable`: What if state was returned instead of assigned?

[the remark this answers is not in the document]
bonndan: changed as proposed. Assessment is a separate thing now with its own API and references to landscape components

src/main/java/de/bonndan/nivio/assessment/Assessable.java
bonndan: this kind of state has been removed competely

src/main/java/de/bonndan/nivio/assessment/Assessable.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Returns the highest status as summary of all {@link StatusValue} and children summaries.
+     *
+     * @return status value, field contains the component identifier, message is the identifier of the highest status value
+     */
+    default StatusValue getOverallStatus() {
+
+        final AtomicReference<StatusValue> summary = new AtomicReference<>();
+        summary.set(new StatusValue(SUMMARY_LABEL, Status.UNKNOWN));
```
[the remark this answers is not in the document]
bonndan: done

PR #130, opened 2020-04-21
refactor home component and add header to all sites

MarvinSchoening:
rename home component and split it into logic and layout, implement navigation component as header, fix styling errors that resulted from those changes


src/main/app/src/Components/LandscapeComponent/LandscapeOverview/LandscapeOverviewLayout.tsx
```diff
@@ -0,0 +1,165 @@
...
+            </Button>
+          </Grid>
+        </Grid>
+      );
+    });
+  }
+
+  return (
+    <div className='homeContainer'>
+      <GenericModal modalContent={modalContent} />
```
bonndan, line 158: Do we still need this here?

bonndan:
:+1: 

I like this refactoring, takes us in the right direction.

src/main/app/src/Components/LandscapeComponent/LandscapeOverview/LandscapeOverviewLayout.tsx
MarvinSchoening: yes to display our log component, we could think about using our modal with a context and only import it once at layout, i'll talk with michelle about it soon

PR #149, opened 2020-06-03
concurrency issues in Test with DirectoryWatcher

mfbieber:
increasing the Thread.sleep() time after changing the file contents allows the application to register the change event (tested on macOS 10.15.4 with openjdk version "11.0.7" 2020-04-14)

### DIFF
--- a/src/test/java/de/bonndan/nivio/input/DirectoryWatcherTest.java
+++ b/src/test/java/de/bonndan/nivio/input/DirectoryWatcherTest.java
@@ -40,7 +40,7 @@ class DirectoryWatcherTest {
 
         Thread.sleep(1000);
         Files.write(tempFile.toPath(), "Hallo".getBytes(), StandardOpenOption.CREATE, StandardOpenOption.APPEND, StandardOpenOption.WRITE);
-        Thread.sleep(SystemUtils.IS_OS_MAC_OSX ? 50000 : 1000); //#147
+        Thread.sleep(SystemUtils.IS_OS_LINUX ? 1000 : 50000); //#147
 
         ArgumentCaptor<FSChangeEvent> captor = ArgumentCaptor.forClass(FSChangeEvent.class);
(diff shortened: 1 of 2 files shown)


mfbieber:
I would suggest not to merge this Pull Request (I only have it here so that we can see the changes more easily for now) and to fix the issue possibly already with https://github.com/dedica-team/nivio/pull/148

src/test/java/de/bonndan/nivio/input/DirectoryWatcherTest.java
```diff
@@ -39,7 +39,7 @@ public void testFileChange() throws IOException, InterruptedException, TimeoutEx
 
         Thread.sleep(1000);
         Files.write(tempFile.toPath(), "Hallo".getBytes(), StandardOpenOption.CREATE, StandardOpenOption.APPEND, StandardOpenOption.WRITE);
-        Thread.sleep(1000);
+        Thread.sleep(50000);
```
bonndan, line 42: Can't we use SystemUtils.IS_OS_MAC_OSX here? Or a lower value? 50 seconds build delay is a bit much for regular CI builds.

mfbieber: Sure. Although I am not sure yet why we are seeing this behaviour. Of course it comes to mind, that it is related to the different file systems. I would still want to check out the build without the Thread.sleep() on e.g. Windows 10 and Windows Server 2016.

You are running the build on your machine on what system? Linux?

In any case, we should investigate the underlying issue, if it is not already fixed with the mentioned PR.

mfbieber:
I saw the second test `testFileChangedIgnoring()` failing on Windows 10 once, then the second and third time it passed.

bonndan:
No crying emojis here?

mfbieber:
:sob::sob::sob:

PR #176, opened 2020-07-03
Replace npm with yarn, update pom.xml, remove unused hexmaps

MarvinSchoening:
A bit cleanup

- frontend now uses yarn instead of npm to make life a bit easier

- update node version in pom.xml to match LTS

- removed hexmaps because we dont render anything in frontend anymore

- update readme informations for new modules/removed old ones and replace npm with yarn


bonndan:
I was able to run yarn and launch nivio's frontend.

PR #178, opened 2020-07-07
Bugfix/169/url helper

BilligsterUser:
URLHelper: update getURL use file.toURI().toURL() instead of 'file:' prefix, add URLHelperTest

Seed: make getLocations return a List<String> , move URL logic to StartupListener.getUrls,remove ESCAPE_AUTHORITY var and toURL function

StartupListener: Update getUrls to create URLs


src/main/java/de/bonndan/nivio/input/Seed.java
```diff
@@ -12,21 +11,30 @@
...
 
+    public static final String DEMO = "DEMO";
     private static final Logger logger = LoggerFactory.getLogger(Seed.class);
+    @Value("${SEED:}")
+    private String seed;
 
-    public static final String DEMO = "DEMO";
+    public Seed() {
 
-    static boolean ESCAPE_AUTHORITY = SystemUtils.IS_OS_WINDOWS;
```
bonndan, line 28: :+1: for removing this completely

src/main/java/de/bonndan/nivio/input/StartupListener.java
```diff
@@ -57,7 +58,19 @@ public void onApplicationEvent(final ApplicationReadyEvent event) {
         List<URL> landscapeDescriptionLocations = new ArrayList<>();
         try {
```
bonndan, line 59: Please consider to move the try-catch inside the loop over seed.getLocations(), so we're able to report single broken URLs and keep the rest.

src/main/java/de/bonndan/nivio/util/URLHelper.java
```diff
@@ -13,38 +13,39 @@
...
+     * <p>
      * Tries to create an URL if a path is given.
      *
      * @param url string url or local path to file
      * @return an URL or null
      */
     @Nullable
     public static URL getURL(String url) {
-        if (url == null) {
+        if (url == null)
```
bonndan, line 30: @basseur will hate that ;)

PR #199, opened 2020-07-23
Feature/198/map relations

### LINKED ISSUE #198: Map relations clickable
To further enhance the user experience, we want the relation paths to be clickable. The click should:

Open a Slider in which:
- [x] Source and Target Relation are clickable and zoom to their map element onClick
- [x] Type attribute should be displayed
- [x] The Panel should transition to the middle of the relation

MarvinSchoening:
- Map Relations are now clickable and open a new slider and highlights the relation
- Reworked some scss abit, because REM shouldn't be used for margin/padding

Closes #198

### DIFF
--- a/src/main/app/src/App.scss
+++ b/src/main/app/src/App.scss
@@ -12,8 +12,8 @@ body {
   .button {
     color: lighten($header-color, 20%);
     padding: 0;
-    border-top: 0.1rem solid $header-color;
-    border-left: 0.3rem solid $header-color;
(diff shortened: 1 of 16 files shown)


MarvinSchoening:
> Please make sure all "jump" buttons have the same type of text, i.e. the fully qualified identifier or preferably the "name".

i changed it to groupName/itemName on both sides, to get the "name" of the relations we would need much effort to get them (2 extra api requests if we target them directly or one if we target their group and some more code to extract the right names)

Issue #14, opened 2019-04-25
Add plugin structure


bonndan:
Change the code (Osgi, Java modules?) to allow extended functionality at deploy/runtime.

For example the dashboard or other types of rendering could be added this way.

bonndan:
Consider using http json communication with plugins (and a contract/schema) instead of tight coupling to java implementations.

bonndan:
Possible ExtensionPoints (pf4j):

* FormatFactory
* Resolver (after last Resolver in Indexer)
* Controller to add own api
* Renderer
* Observer (input changes)
* KPIs

PR #204, opened 2020-07-24
Rendering without JGraphX

### LINKED ISSUE #193: Landscape layout changes after reindexing
A layout can completeley change (e.g. from more horizontal alignment to vertical) if a source reference changes content (e.g. name  is changed from "Wordpress" to "Wordpress XYZ") and the landscape is reindex.

* find out why and fix if possible (hint: see initial random seed in JGraphX, or maybe sequence of items)

bonndan:
This escalated quickly:

* removes jgraphx dependency from nivio, now it is visible what is happening during layout etc (keeping the organic layout code from jgraphx with some modifications)
* renders groups as hex areas, not as dumb rectangles (to be drawn as Civilisation-like border later)
* ensures that similar input leads to similar output (no more flickering or unexpected position changes) closes #193 

- [x] please find out why the os x build runs into a timeout


src/main/java/de/bonndan/nivio/output/map/svg/SVGRenderer.java
```diff
@@ -1,33 +1,113 @@
...
+
+            groupBounds.getChildren().forEach(itemBounds -> {
+                LOGGER.debug("original item pos {} {}", itemBounds.getX(), itemBounds.getY());
+                itemBounds.setX(itemBounds.getX() + groupBounds.getX());
+                itemBounds.setY(itemBounds.getY() + groupBounds.getY());
+                LOGGER.debug("item pos with group offset: {} {}", itemBounds.getX(), itemBounds.getY());
+
+                Item item = (Item) itemBounds.getComponent();
+                itemBounds.setColor(group.getColor());
+                itemBounds.setIcon(localServer.getIconUrl(item).toString());
```
BilligsterUser, line 73: This will link the image from 'http://localhost:8080/icons/service.png'
maybe it is better to embed the image base64 encoded.
I think this is also the reason why 'https://github.com/dedica-team/nivio/actions/runs/183402402' fails.

bonndan: fixed

src/main/java/de/bonndan/nivio/output/layout/OrganicLayouter.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class OrganicLayouter implements Layouter<LayoutedComponent> {
+
+    public OrganicLayouter() {
+    }
+
+    @Override
+    public LayoutedComponent layout(LandscapeImpl landscape) {
+
+        Map<String, SubGraph> subgraphs = new LinkedHashMap<>();
```
mfbieber, line 22: Do we need a LinkedHashMap, or could it be a HashMap? (I don't think that we have memory problems with nivio, just to understand)

src/main/java/de/bonndan/nivio/output/map/PNGRenderCache.java
```diff
@@ -71,16 +69,15 @@ public PNGRenderCache(MapFactory<mxGraph, mxCell> mapFactory, MapStyleSheetFacto
      * @return the svg as string, uncached
      */
     public String getSVG(LandscapeImpl landscape) {
-        JGraphXRenderer jGraphXRenderer = new JGraphXRenderer();
-        RenderedArtifact<mxGraph, mxCell> render = jGraphXRenderer.render(landscape);
-        mapFactory.applyArtifactValues(landscape, render);
+        OrganicLayouter layouter = new OrganicLayouter();
```
mfbieber, line 72: Why don't we use the interface `Layouter`?

src/main/java/de/bonndan/nivio/output/map/PNGRenderCache.java
```diff
@@ -93,23 +90,17 @@ public void onApplicationEvent(ProcessingFinishedEvent processingFinishedEvent)
...
     private byte[] asByteArray(LandscapeImpl landscape) {
-        TranscoderInput input_svg_image = new TranscoderInput(new ByteArrayInputStream(getSVG(landscape).getBytes()));
-        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        TranscoderOutput transcoderOutput = new TranscoderOutput(outputStream);
+        TranscoderInput input_svg_image = new TranscoderInput(
+                new ByteArrayInputStream(getSVG(landscape).getBytes())
+        );
         Transcoder transcoder = new PNGTranscoder();
-        try {
+        try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
```
mfbieber, line 97: @BilligsterUser: nice that you fixed that error! Could you also write a test that fails with the old implementation and now passes?

src/main/java/de/bonndan/nivio/output/layout/AllGroupsGraph.java
```diff
@@ -0,0 +1,111 @@
...
+            items.addAll(groupItem.getItems());
+        });
+        LOGGER.debug("Group node sequence: {}", groupNodes);
+
+        addVirtualEdgesBetweenGroups(items);
+
+        layout = new FastOrganicLayout(new ArrayList<>(groupNodes.values()));
+        //layout.setDebug(true);
+
+        Optional.ofNullable(landscape.getConfig().getJgraphx().getMaxIterations())
```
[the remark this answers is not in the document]
bonndan: yes, and this also requires updating the documentation.

src/main/java/de/bonndan/nivio/input/GroupQueryResolver.java
```diff
@@ -30,5 +31,11 @@ public void process(LandscapeDescription input, LandscapeImpl landscape) {
             }
             groupDescription.getContains().forEach(condition -> group.getItems().addAll(landscape.getItems().query(condition)));
         });
+
+        //cleanup to ensure every items has the group identifier
+        landscape.getItems().stream().forEach(item -> {
+            Group group = landscape.getGroup(item.getGroup()); //if group is empty, COMMON is returned
```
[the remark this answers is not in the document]
bonndan: The input DTOs might not have the group reference, and all following resolvers might fail to find or set a group. So this is a fallback. I'll add this explanation to the code.

src/main/java/de/bonndan/nivio/input/GroupQueryResolver.java
```diff
@@ -30,5 +31,11 @@ public void process(LandscapeDescription input, LandscapeImpl landscape) {
...
+            item.setGroup(group.getIdentifier());
```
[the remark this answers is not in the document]
bonndan: The method landscape.getGroup() has some magic to be non-nullable, so here we ensure the item.group field is set to something that really points to a group. Could make an example how Landscape.updateGroupsOnItems might be used?

src/main/java/de/bonndan/nivio/output/dld4e/Dld4eRenderer.java
```diff
@@ -14,6 +15,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+@Deprecated //this is both layouter and renderer
```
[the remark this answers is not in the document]
bonndan: This was an early approach to use dld4e (see https://github.com/bonndan/dld4e-headless) to draw landscapes, but it lacked automatic layouting. It was kept in case it could become a plugin (#14) or somebody might be interested to use it with the automatic layouting. For now I'd say there is no replacement.

src/main/java/de/bonndan/nivio/output/dld4e/Dld4eRenderer.java
```diff
@@ -27,10 +29,10 @@
     private final Icons icons = new Icons();
 
     @Override
-    public String render(LandscapeImpl landscape) {
+    public String render(LayoutedComponent landscape) {
 
-        landscape.getItems().stream().forEach(this::addService);
-        landscape.getItems().stream().forEach(this::addLinks);
+        List<Item> items = new ArrayList<>();
```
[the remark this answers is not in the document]
bonndan: I would not repair that old code.

src/main/java/de/bonndan/nivio/output/layout/GroupConnections.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+
+        if (a.equals(b))
+            return false;
+
+        boolean hasLink = groupConnections.stream()
+                .anyMatch(pair ->
+                        (pair.getKey().equals(a) && pair.getValue().equals(b))
+                               // || (pair.getKey().equals(b) && pair.getValue().equals(a))
```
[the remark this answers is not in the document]
bonndan: ... because we now need the inverse direction, too. JGraphX handled that before it was removed.

src/main/java/de/bonndan/nivio/output/layout/OrganicLayouter.java
bonndan: It is used to ensure that insertion sequence is maintained. Otherwise layout might receive the items in random order, leading to random results.

src/main/java/de/bonndan/nivio/output/layout/OrganicLayouter.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *
+ */
+public class OrganicLayouter implements Layouter<LayoutedComponent> {
+
+    public OrganicLayouter() {
+    }
+
+    @Override
+    public LayoutedComponent layout(LandscapeImpl landscape) {
```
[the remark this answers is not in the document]
bonndan: As written in our chat, the LandscapeImpl is the real graph, and the Landscape interface might not make too much sense anymore.

src/main/java/de/bonndan/nivio/input/GroupQueryResolver.java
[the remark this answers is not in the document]
mfbieber: Ok, so it is happening inside of the `Group`, now I see it.

src/main/java/de/bonndan/nivio/output/dld4e/Dld4eRenderer.java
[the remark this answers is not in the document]
mfbieber: So, it is deprecated but we have no replacement? 

I still don't understand what to do about the Dld4eRenderer :) If it is deprecated it means that we don't want it to be used anymore. We should add a comment on what to use instead.

Otherwise, if we want to keep it, it should not be deprecated.

src/main/java/de/bonndan/nivio/output/layout/OrganicLayouter.java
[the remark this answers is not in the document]
mfbieber: Ok, if it does not make sense anymore, how about deleting it?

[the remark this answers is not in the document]
bonndan: Yes, but that takes some days of refactoring. It's not adding business value and not scope of this feature. I've added #212

src/main/java/de/bonndan/nivio/output/dld4e/Dld4eRenderer.java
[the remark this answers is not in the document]
bonndan: Actually I am too lazy to reimplement dld4e rendering with with new layout, although it might not take too long. I know it's in the repo, but I'm still afraid it will be forgotten if we'd delete it. If we're strict then it should be deleted.

src/main/java/de/bonndan/nivio/input/GroupQueryResolver.java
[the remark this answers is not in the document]
mfbieber: Yes, I saw that it uses that `Group.COMMON` and thanks for your explanation! I am still concerned about the mutability of `Item` and of the way we expose the internals of the objects.

So if your example means that you would agree to a refactoring (because you call it `Landscape.updateGroupsOnItems`): yes, please and I'll do the refactoring if you like.

I am still wondering why the identifier can not be set on the item right from the beginning? Why exactly isn't it possible to have such (important) fields immutable right from the beginning when we create an instance of the object?

mfbieber:
Does `GroupQueryResolver` already have a test? I only saw that `GroupResolver` is tested.

src/main/java/de/bonndan/nivio/output/dld4e/Dld4eRenderer.java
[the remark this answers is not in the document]
mfbieber: Mhm, well you decide. My opinion is: get rid of it. If it is important, we have the VCS to get the code back ;)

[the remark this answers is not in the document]
mfbieber: Otherwise, it wasn't important 😸

src/main/java/de/bonndan/nivio/input/GroupQueryResolver.java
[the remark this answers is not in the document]
bonndan: You are right about the immutability, although some things cannot be set right from the start. And if we're strict about immutability and side effect free programming, we end in reference update hell (this is still a graph with lots of references).

My idea was that all graph components are value objects without much business logic. The logic was meant to reside in the resolvers as far as possible. So we have small units (resolvers) for one dedicated purpose and small value objects. 
Obviously the landscape.getGroup() method does too much, and I'd rather go the other way and refactor even more logic out of the graph components.

src/main/java/de/bonndan/nivio/output/map/PNGRenderCache.java
bonndan: I've pushed a change. Please see if you agree with it. This code is likely to change later when there are more than just one layouter and render implementations.

src/main/java/de/bonndan/nivio/output/layout/AllGroupsGraph.java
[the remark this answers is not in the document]
bonndan: done. also separated configs into two: one for group layout, one for layout within groups

src/main/java/de/bonndan/nivio/output/dld4e/Dld4eRenderer.java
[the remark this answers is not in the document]
bonndan: Removed it. :crying_cat_face:

src/main/java/de/bonndan/nivio/input/GroupQueryResolver.java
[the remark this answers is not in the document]
mfbieber: Ok, I get that about the resolvers and the value object - that approach is fine. Since I lack the overview of how the value objects are used, I was concerned about the objects having an invalid state (in a field such as an identifier). If only a designated resolver manages them, and the logic resides inside of that - that should be fine.

src/main/java/de/bonndan/nivio/output/map/PNGRenderCache.java
mfbieber: Yes, if instantiation in the constructor makes sense because we don't need a fresh layouter each time we call `getSVG()`, sure, why not. This was just intended as a question to understand how we are using interfaces, and so on :)

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
```diff
@@ -0,0 +1,97 @@
...
+        LandscapeDescription input = new LandscapeDescription();
+        GroupDescription groupDescription = new GroupDescription();
+        groupDescription.setIdentifier("groupIdentifier");
+        groupDescription.setContains(List.of("itemidentifier"));
+        input.getGroups().put("group", groupDescription);
+
+        groupResolver.process(input, landscape);
+
+        assertThat(landscape.getItems().stream().count()).isEqualTo(1L);
+        assertThat(landscape.getGroup("groupIdentifier").getItems()).containsExactly(item);
```
mfbieber, line 51: This list contains the same `Item` twice. This is because the items are just added to the modifiable Collection in `Group`. 

I would refactor this to add a method `addItem` to `Group` that does not add items twice (or use a Set) and to have an unmodifiable Collection instead.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
```diff
@@ -0,0 +1,97 @@
...
+        landscape.setItems(Set.of(item));
+        landscape.addGroup(new Group("groupIdentifier"));
+
+        LandscapeDescription input = new LandscapeDescription();
+        GroupDescription groupDescription = new GroupDescription();
+        groupDescription.setIdentifier("groupIdentifier");
+        groupDescription.setContains(List.of("itemidentifier"));
+        input.getGroups().put("group", groupDescription);
+
+        groupResolver.process(input, landscape);
```
mfbieber, line 48: This shows what has to be done to have valid objects (instantiation and setting of things), only to not have a NPE. 

I am not even sure if this is how the objects are intended to be used 😞, which means that somehow it is not straightforward and thus potentially unsafe.

I also did not really understand why we have `Group` and `GroupDescription` both implementing `GroupItem`. There are several casts to these objects, which are really hard to get rid of.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
```diff
@@ -0,0 +1,97 @@
...
+        Item item = new Item();
+        item.setIdentifier("itemIdentifier");
+        item.setGroup("groupIdentifier");
+        landscape.setItems(Set.of(item));
+        landscape.addGroup(new Group("groupIdentifier"));
+
+        LandscapeDescription input = new LandscapeDescription();
+        GroupDescription groupDescription = new GroupDescription();
+        groupDescription.setIdentifier("groupIdentifier");
+        groupDescription.setContains(List.of("itemidentifier"));
```
mfbieber, line 45: The query condition has to be the lowercased identifier. Is this intentional? See the other test below.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+        LandscapeDescription input = new LandscapeDescription();
+        GroupDescription groupDescription = new GroupDescription();
+
+        input.getGroups().put("group", groupDescription);
+
+        groupResolver.process(input, landscape);
+
+        assertThat(landscape.getItems().stream().count()).isEqualTo(1L);
+        assertThat(landscape.getGroups().values().size()).isEqualTo(1);
```
mfbieber, line 93: This aims for the las stream of the `process()` method. The groups are not updated on the landscape, only `getGroup(somethingInvalid)` returns `Group.COMMON`. Is this intentional?

src/main/java/de/bonndan/nivio/model/LandscapeImpl.java
```diff
@@ -162,12 +162,13 @@ public void addGroup(Group g) {
...
     public Group getGroup(String group) {
-        if (StringUtils.isEmpty(group))
+        if (StringUtils.isEmpty(group)) {
             group = Group.COMMON;
+        }
         GroupItem groupItem = groups.get(group);
         if (groupItem == null) {
             processLog.warn("Group " + group + " was not found, returning COMMON");
-            groupItem = groups.get(Group.COMMON);
+            groupItem = new Group(Group.COMMON);
```
mfbieber: `groups` can be an empty Map, so null was returned before. Is my change how it was intended to be? 

If yes, do we need that `COMMON` logic inside of `Group`?

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
bonndan: As you suggested we are now using a Set.

bonndan: The unmodifiable Collection thing is a bit too much at the moment. We are constantly updating the collections in the indexing phase, and having side-effect free copies of copies of copies does not seem worth the effort to me.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
bonndan: Please see LandscapeFactory, it now sets up a proper landscape. We might think about more safety measures though.

Concerning Group and GroupDescription: #212 .

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
bonndan: Good point. We must make sure that querying is not case-sensitive.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
bonndan: Has been refactored.

src/main/java/de/bonndan/nivio/model/LandscapeImpl.java
bonndan: Has been refactored.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
mfbieber: Ok, I guess most of my suggestions would require some other patterns in our code, which can be hard to implement now into the existing code base with it's patterns already in place.

src/main/java/de/bonndan/nivio/model/LandscapeImpl.java
mfbieber: Alright, but the resulting landscape still does not have a group, is this intentional?

mfbieber: If you enable the `process_containsCommonGroupIfGroupWasEmpty()` in `GroupQueryResolverTest`, you should see:
````

java.util.NoSuchElementException: No value present

	at java.base/java.util.Optional.get(Optional.java:148)
	at de.bonndan.nivio.input.GroupQueryResolver.process(GroupQueryResolver.java:36)
...
````

mfbieber: I added a (passing) Test that does exactly the same as the other mentioned here, but that uses a Landscape from the LandscapeFactory. 

For a safety mechanism to use the Landscape correctly: isn't it possible to enforce the same things that the LandscapeFactory does via the Landscape's constructor?

bonndan: Yes :)

mfbieber: An that `get()` on the optional in `GroupQueryResolver` should have a preceding `ifPresent()` check and an `orElse()` handling :)

bonndan: I am working on it.

mfbieber: I was testing with:
````
     public LandscapeImpl(@Pattern(regexp = LandscapeItem.IDENTIFIER_VALIDATION) String identifier) {
        this.identifier = identifier;
    }

    @JsonIgnore
    @Override
    public Map<String, GroupItem> getGroups() {
        if (groups.isEmpty()) {
            addGroup(new Group(Group.COMMON));
        }
        return groups;
    }
````

But then `AllGroupsLayoutTest.testWithARelation()` and `GroupResolverTest.testBlacklistOnGroupsWithRegex()` fails (aswell as tests in `OrganicLayouterTest`, but these due to the `.get()` on the Optional I guess).

So the groups map is supposed to be empty sometimes?

mfbieber: Ok, I see your changes - nice 😻 

`SubLayoutTest.testWithARelation()` is failing though 😿

bonndan: Where is this code? However, I've refactoded the landscape constructor and such code is not necessary anymore.

src/test/java/de/bonndan/nivio/input/GroupQueryResolverTest.java
bonndan: Yes. Please feel free to open tickets for enhancement proposals and simply extend #212

src/main/java/de/bonndan/nivio/model/LandscapeImpl.java
mfbieber: I didn't push it. I knew you were working on the same thing, so I was just checking that out locally.

src/main/java/de/bonndan/nivio/output/map/PNGRenderCache.java
BilligsterUser: 83ba1a4

PR #211, opened 2020-07-29
fix URLHelper.splitQuery & add test

BilligsterUser:


src/test/java/de/bonndan/nivio/util/URLHelperTest.java
```diff
@@ -37,4 +38,28 @@ public void test_getURL_relativePath() {
         assert url != null;
         assertEquals("file:" + root + "/src/test/resources/example/example_templates.yml", url.toString());
     }
+
+    @Test
+    void test_splitQuery() throws MalformedURLException {
+        URL url = new URL("http://192.168.99.100:8080?namespace=default&groupLabel=release");
+        Map<String, String> expected = Map.of("namespace", "default", "groupLabel", "release");
```
bonndan, line 45: What type of Map is used in Map.of()? Is the sequence guaranteed? Or does it not matter in assertEquals?

PR #215, opened 2020-08-05
Add format test in CI and remove husky

MarvinSchoening:
Added format CI to test if everything is formatted properly (only frontend currently). Tests are already performed in our OS check steps.


MarvinSchoening:
squash commits on merge!

PR #235, opened 2020-08-17
Omit dependency download logs during Maven test run

Matthimatiker:
Ensures that the download progress of dependencies is not shown in the GitHub Actions output.

Additionally, both GitHub Actions jobs use the same checkout step version now.


bonndan:
Awesome! I've always tried to omit the logs just by using -B..

Issue #231, opened 2020-08-15
Add ArchUnit tests


bonndan:
See https://github.com/TNG/ArchUnit

Desired: Prototypical integration.

- [x] add a standard tests for Java (solved by https://github.com/dedica-team/nivio/pull/252)
  - [x] Rule: Do not write to StdOut/StdErr
- [x] add a standard tests for Spring Boot
  - [x] Rule: Do not define Beans in Application class

- [x] see if tests can be replaced with https://github.com/societe-generale/arch-unit-maven-plugin

PR #252, opened 2020-09-06
[#231] Add ArchUnit tests

BilligsterUser:
See https://github.com/TNG/ArchUnit

Desired: Prototypical integration.

- add a standard tests for Java
  - [x] Rule: Do not write to StdOut/StdErr
- add a standard tests for Spring Boot
  - [ ] Rule: Do not define Beans in Application class

ItemIndex: Use LOGGER.warn() instead of ex.printStackTrace()


bonndan (changes requested):
Good, this solves part 1 of the ticket. 👍

src/main/java/de/bonndan/nivio/model/ItemIndex.java
```diff
@@ -241,7 +245,7 @@ public int indexForSearch() {
             ireader.close();
             return facets.getAllDims(10);
         } catch (IOException e) {
-            e.printStackTrace();
+            LOGGER.warn("Unable to get the facets for the given query error: ", e);
```
Matthimatiker, line 248: 👍

src/main/java/de/bonndan/nivio/model/ItemIndex.java
```diff
@@ -259,8 +263,8 @@ public int indexForSearch() {
 
         List<Document> documents = new ArrayList<>();
         // Iterate through the results:
-        for (int i = 0; i < hits.length; i++) {
-            Document hitDoc = isearcher.doc(hits[i].doc);
+        for (ScoreDoc hit : hits) {
```
Matthimatiker, line 266: More readable 👍

src/test/java/de/bonndan/nivio/ArchTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import static com.tngtech.archunit.library.GeneralCodingRules.NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS;
+
+class ArchTest {
+    @Test
+    void do_not_access_STD_streams() {
+        JavaClasses classes = new ClassFileImporter()
+                .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_JARS)
+                .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
+                .importClasspath();
```
Matthimatiker, line 16: You could move the classes into a test attribute, even a static one.

Usually, the classes are used by many (if not all) ArchUnit tests and collecting the data from byte code is expensive.

Matthimatiker (commented):
Good start 👍  
Thanks for cleaning up the code you come across.

src/test/java/de/bonndan/nivio/ArchTest.java
mfbieber: I took over to do this to merge this PR. I need ArchUnit for Issue 317!

bonndan (approved):

Issue #212, opened 2020-07-29
Cleaner Code: make landscape graph less error-prone and more consistent


bonndan:
The components of the landscape graph are highly mutable, what introduces unpredictable state at some points.

- [ ] change the constructors of the landscape components (Landscape, Group, Item) to contain the required fields (like identifier and such)
- [ ] look into immutable objects, especially the ones in the graph

FYI #317 was part of this ticket before

PR #263, opened 2020-09-09
260 duplicate items

### LINKED ISSUE #260: [Bug] Duplicated items on rename
If you rename an item while nivio is running, it sometimes creates duplicates of existing items.

Steps for reproducing:
1. Start backend
2. go to test\resources\example\services\wordpress.yml
3. change the name of Demo Blog to something else (not the identifier)

bug

bonndan:
closes #260 

The problem is that Item.hashCode is dependent on the "group" field value, which had been changed after items were added to a Set. Usually Sets only contain unique items, but the item group was changed after having been added to the collection, so the equality to duplicates was not given anymore.

In addition, Group.getItems now returns an immutable copy and an addItem method has been added which enforces correct item group field values.

This is a step towards #212 , but does not eliminate all mutability.

### DIFF
--- a/src/main/java/de/bonndan/nivio/model/Group.java
+++ b/src/main/java/de/bonndan/nivio/model/Group.java
@@ -98,7 +98,7 @@ public class Group implements GroupItem, Labeled, Assessable {
      * @return immutable copy
      */
     public Set<Item> getItems() {
-        return Set.copyOf(items);
+        return Collections.unmodifiableSet(items);
     }
 
     @Override
--- a/src/main/java/de/bonndan/nivio/model/Group.java
+++ b/src/main/java/de/bonndan/nivio/model/Group.java
@@ -20,7 +20,7 @@ public class Group implements GroupItem, Labeled, Assessable {
      * Items belonging to this group. Order is important for layouting (until items are ordered there).
      */
     private final Set<Item> items = new LinkedHashSet<>();
-    private String identifier;
+    private final String identifier;
     private String owner;
     private String description;
     private String contact;
@@ -29,16 +29,15 @@ public class Group implements GroupItem, Labeled, Assessable {
     private String landscapeIdentifier;
 
     public Group(String identifier) {
-        setIdentifier(identifier);
+        if (StringUtils.isEmpty(identifier)) {
+            throw new IllegalArgumentException("Group identifier must not be empty");
+        }
+        this.identifier = identifier;
     }
 
     @Override
     public String getIdentifier() {
         return identifier;
-    }
-
-    public void setIdentifier(String identifier) {
-        this.identifier = StringUtils.isEmpty(identifier) ? COMMON : identifier;
     }
 
     @Override
@@ -93,8 +92,13 @@ public class Group implements GroupItem, Labeled, Assessable {
         return links;
     }
 
+    /**
+     * Returns an immutable copy of the items.
+     *
+     * @return immutable copy
+     */
     public Set<Item> getItems() {
-        return items;
+        return Set.copyOf(items);
     }
 
     @Override
@@ -141,5 +145,23 @@ public class Group implements GroupItem, Labeled, Assessable {
         return "Group{" +
                 "identifier='" + identifier + '\'' +
                 '}';
+    }
+
+    /**
+     * Adds an item to this group.
+     *
+     * @param item the item to add.
+     * @throws IllegalArgumentException if the item group field mismatches
+     */
+    public void addItem(Item item) {
+        boolean canAdd = item.getGroup() == null || (item.getGroup().equals(identifier));
+
+        if (canAdd) {
+            item.setGroup(identifier);
+            items.add(item);
+            return;
+        }
+
+        throw new IllegalArgumentException(String.format("Item group '%s' cannot be added to group '%s'", item.getGroup(), identifier));
     }
 }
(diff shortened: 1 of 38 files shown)


BilligsterUser (approved):

BilligsterUser (approved):

src/main/java/de/bonndan/nivio/model/Group.java
```diff
@@ -93,8 +92,13 @@ public void setColor(String color) {
...
     }
 
+    /**
+     * Returns an immutable copy of the items.
+     *
+     * @return immutable copy
+     */
     public Set<Item> getItems() {
-        return items;
+        return Set.copyOf(items);
```
Matthimatiker, line 101: You could also wrap it with `Collections.unmodifiableSet()` as modification would not make sense anymore.

src/main/java/de/bonndan/nivio/model/Group.java
```diff
@@ -142,4 +146,22 @@ public String toString() {
...
                 '}';
     }
+
+    /**
+     * Adds an item to this group.
+     *
+     * @param item the item to add.
+     * @throws IllegalArgumentException if the item group field mismatches
+     */
+    public void addItem(Item item) {
```
Matthimatiker, line 156: :+1:

src/main/java/de/bonndan/nivio/model/Item.java
```diff
@@ -18,7 +18,7 @@
 
     @NotNull
     @Pattern(regexp = LandscapeItem.IDENTIFIER_VALIDATION)
-    private String identifier;
+    private final String identifier;
```
Matthimatiker, line 21: :+1:

src/main/java/de/bonndan/nivio/model/Item.java
```diff
@@ -227,7 +236,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(toString());
+        return Objects.hash(identifier);
```
Matthimatiker, line 239: :+1:

Matthimatiker: Shouldn't equals also check for the `identifier` now?

src/test/java/de/bonndan/nivio/assessment/kpi/LifecycleKPITest.java
```diff
@@ -24,8 +24,7 @@ public void setup() {
 
     @Test
     public void green() {
-        Item item = new Item();
-        item.setIdentifier("foo");
+        Item item = new Item(null, "foo");
```
Matthimatiker, line 27: Adding a constructor for this case might increase readability of item construction, as the reader does not have to reason about `null`.

src/test/java/de/bonndan/nivio/model/GroupTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Test
+    void doesNotAllowEmptyIdentifier() {
+        assertThrows(Exception.class, () -> new Group(""));
+    }
+
+    @Test
+    void getItemsIsImmutable() {
+        Group g = new Group("foo");
+        assertThrows(Exception.class, () -> g.getItems().add(new Item("a", "b")));
```
Matthimatiker, line 17: I am a bit surprised that this throws an exception.

I would have expected that the returned `Set` (but not the original one) is modified.

Matthimatiker (approved):

src/main/java/de/bonndan/nivio/model/Group.java
bonndan: Set.of is the same, but unmodifiableSet is better to understand, so I'll change it.

src/test/java/de/bonndan/nivio/model/GroupTest.java
bonndan: The point is that group "foo" would not allow that an item of group "a" can be added (item.group cannot be changed anymore). 

Messing around with strings should be replaced in #212

src/test/java/de/bonndan/nivio/assessment/kpi/LifecycleKPITest.java
bonndan: In productive code the constructor is only used in ItemFactory, and there both args are given.

src/main/java/de/bonndan/nivio/model/Item.java
bonndan: I'd rather say no, because equality is given if FullyIdentifiedQualifier is equal, and this means that identifier AND group are equal (group comparison is a bit more complex, and that is encapsulated in FQI)

Matthimatiker: I don't know enough about the internals to decide, which equality is right.

`equals() ` that uses `toString()` looked a bit unusual.
We just have to make that `hashCode()` and `equals()` fit together: If 2 objects are equal, they always *must* return the same hash code.

src/test/java/de/bonndan/nivio/model/GroupTest.java
Matthimatiker: I was mainly surprised by the `Set.copyOf()` behavior.

PR #273, opened 2020-09-13
248 observe source reference

bonndan:
This changes enables nivio to react on input changes depending on the format. Previously only changes in files could be detected, but not changes in other input types like rancher, k8s etc.

Follow up: #274


src/main/java/de/bonndan/nivio/util/URLHelper.java
```diff
@@ -23,40 +25,40 @@ public static boolean isLocal(URL url) {
...
         if (url == null) {
-            return null;
+            return Optional.empty();
         }
         try {
-            return new URL(url);
+            return Optional.of(new URL(url));
         } catch (MalformedURLException e) {
             File file = new File(url);
             if (file.exists() && !url.startsWith("file:")) {
```
Matthimatiker, line 38: What does this mean? Can be a file but must *not* be a ``file://`` URL?

src/main/java/de/bonndan/nivio/util/URLHelper.java
```diff
@@ -23,40 +25,40 @@ public static boolean isLocal(URL url) {
...
+    public static Optional<URL> getParentPath(String url) {
+
+        Optional<URL> url1 = getURL(url);
+        if (url1.isEmpty()) {
+            return Optional.empty();
         }
+
         try {
-            return url1.toURI().resolve(".").toURL();
+            return Optional.of(url1.get().toURI().resolve(".").toURL());
```
Matthimatiker, line 59: Could also be:

    url1.map(url -> url.toURI().resolve(".").toURL());

That would make the previous empty check unnecessary.

src/test/java/de/bonndan/nivio/input/InputFormatHandlerFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    public void compose2() {
+
+        InputFormatHandler factory = formatFactory.getInputFormatHandler(
+                new SourceReference(null, "docker-compose-v2"),
+                new LandscapeDescription()
+        );
+
+        assertTrue(factory instanceof InputFormatHandler);
+        assertTrue(factory instanceof InputFormatHandlerCompose2);
```
Matthimatiker, line 58: Already using AssertJ?

Using

     assertThat(factory).isInstanceOf(InputFormatHandlerCompose2.class)

would provide a way better message in case of failure.

src/test/java/de/bonndan/nivio/input/SourceReferencesResolverTest.java
```diff
@@ -35,8 +34,11 @@
     public void setup() {
         log = new ProcessLog(LoggerFactory.getLogger(SourceReferencesResolver.class));
         sourceReferencesResolver = new SourceReferencesResolver(
-                new ItemDescriptionFormatFactory(
-                        new ArrayList<ItemDescriptionFactory>(Arrays.asList(ItemDescriptionFactoryNivio.forTesting(), ItemDescriptionFactoryCompose2.forTesting()))
+                new InputFormatHandlerFactory(
+                        new ArrayList<>(Arrays.asList(
+                                new InputFormatHandlerNivio(new FileFetcher(new HttpService())),
```
Matthimatiker, line 39: Isn't there the risk that creating ``HttpService`` in this test makes real HTTP requests or spins up a connection pool?
The ``HttpService`` or ``FileFetcher`` could be mocked to ensure that this does not happen.

src/test/java/de/bonndan/nivio/observation/FileSourceReferenceObserverTest.java
```diff
@@ -0,0 +1,69 @@
...
+        String source = getRootPath() + "/src/test/resources/example/example_env.yml";
+        File file = new File(source);
+
+        LandscapeDescription description = landscapeDescriptionFactory.fromYaml(file);
+
+        SourceReference reference = description.getSourceReferences().get(0);
+        reference.setUrl("./services/wordpress.yml");
+
+        FileFetcher mocked = mock(FileFetcher.class);
+        when(mocked.get(any(SourceReference.class), any(URL.class))).thenAnswer(invocationOnMock -> String.valueOf(Math.random()));
```
Matthimatiker, line 52: Avoiding random numbers would make this test more deterministic.

Doesn't this test fail when the same number is generated 2 times in a row?

src/test/java/de/bonndan/nivio/observation/ObserverRegistryTest.java
```diff
@@ -4,77 +4,54 @@
...
-        assertEquals(landscape.getIdentifier(), observedLandscapes.iterator().next());
-
-        verify(urlObserverFactory).getObserver(any(URL.class));
+        observerPoolFactory = mock(LandscapeObserverPoolFactory.class);
+        observerRegistry = new ObserverRegistry(observerPoolFactory, landscapeDescriptionFactory, publisher);
     }
 
     @Test
-    public void willRegardRelativePaths() {
+    public void onApplicationEvent() {
```
Matthimatiker, line 41: A more descriptive test name would be helpful here.

Matthimatiker (approved):
I only had several small remarks.

Considering how often the ``FileFetcher`` is instantiated with a real ``HttpClient``  in the tests, I would recommend to add a testing version of it in the future that simulates missing network access.
That would reduce the setup code and ensure that no HTTP requests are triggered by tests.

src/main/java/de/bonndan/nivio/input/InputFormatHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Returns an observer for the source reference.
+     *
+     * @param reference the source reference with a format
+     * @param baseUrl   the url of the landscape description
+     * @return observer that can handle the format or null if no observer is available
+     */
+    @Nullable
+    InputFormatObserver getObserver(SourceReference reference, @Nullable URL baseUrl);
```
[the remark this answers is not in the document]
bonndan: Returning null is OK for now since we don't have k8s/rancher  (see #275 ) observers. Later, however, we should have observers for every input type and returning null should be a rare edge case.

src/main/java/de/bonndan/nivio/input/InputFormatHandlerFactory.java
```diff
@@ -3,22 +3,23 @@
...
 @Service
-public class ItemDescriptionFormatFactory {
+public class InputFormatHandlerFactory {
 
-    private final Map<ItemDescriptionFactory, List<String>> factoryListMap = new ConcurrentHashMap<>();
+    private final Map<InputFormatHandler, List<String>> factoryListMap = new ConcurrentHashMap<>();
 
-    public static ItemDescriptionFormatFactory with(ItemDescriptionFactory factory) {
-        return new ItemDescriptionFormatFactory(new ArrayList<>(Collections.singletonList(factory)));
+    public static InputFormatHandlerFactory with(InputFormatHandler factory) {
```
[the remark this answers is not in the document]
bonndan: 👍

src/test/java/de/bonndan/nivio/input/InputFormatHandlerFactoryTest.java
bonndan: 😁

src/main/java/de/bonndan/nivio/util/URLHelper.java
bonndan: 👍

src/test/java/de/bonndan/nivio/observation/FileSourceReferenceObserverTest.java
bonndan: I've been too lazy to look up this simple code:

`when(mocked.get(any(SourceReference.class), any(URL.class))).thenReturn("foo", "bar")`

src/test/java/de/bonndan/nivio/input/SourceReferencesResolverTest.java
bonndan: I've replaced everything with mocks but one.

PR #296, opened 2020-09-20
Plant UML diagram of Nivio backend architecture

mfbieber:


backend-architecture-api.puml
```diff
@@ -0,0 +1,178 @@
...
+
+class ItemDescriptionFormatFactory {
+    +ItemDescriptionFactory getFactory()
+}
+
+class Indexer {
+Triggered by ApplicationEvents fired by observing changes in files through the
+FileSourceReferenceObserver
+--
+    -LandscapeRepository landscapeRepo
```
mfbieber, line 100: Is this the only one who triggers the Indexer?

So the Observers are only for registering file changes, right? And then an ApplicationEvent is fired, which in turn triggers the Indexer?

Did I miss anything else? Are the ApplicationEvents used for anything else?

bonndan: The only way to trigger indexing is indeed an ApplicationEvent (IndexEvent). This happens at application start via SEED variable content, at observed changes or via API (not sure if that still works).

BilligsterUser (approved):
+1

readme.md
```diff
@@ -81,4 +81,30 @@ Clone nivio, build and run a Docker image:
...
+
+This `LandscapeDescription` has to be enriched with the `ItemDescription` for all items in the landscape.
+This is managed by the `InputFormatHandler`, which are able to read several input formats such as e.g. kubernetes files, or the
+nivio description format. Access to these handlers is managed by the `InputFormatHandlerFactory`.
+
+To actually create a landscape, the `Indexer` is used. This is able to compute the landscape graph from a `LandscapeDescription`.
+The `Indexer` uses several `Resolver` to resolve groups and item relations in landscapes and e.g. the appearance of the graph.
+
+The `Indexer` can be triggered either directly through the `ApiController` to index or reindex a landscape, or it is triggered
+by an observer mechanism on files. These are the files located under the path provided through the `SEED` environment variable.
```
mfbieber, line 109: @bonndan : we discussed above that the `Indexer` is triggered only by an `ApplicationEvent`. In the `ApiController` though, the endpoints `indexLandscape()` and `reindex()` both also trigger the `Indexer`.

Is my description correct?

bonndan: we should maybe conside limiting the triggering to events, this way the frontend is notified of the event, too.

bonndan: I'll open a new ticket.

PR #311, opened 2020-09-26
307 fix group areas

### LINKED ISSUE #264: PathFinder does not always use best path
image

### LINKED ISSUE #307: Fix gaps in group areas
The currently algorithm to create group areas works by collecting hex tiles which are neighbours of items and closing gaps. If two item lie too far apart, i.e. the gap is too big, separate islands are created:

image

- [x] items in the same group without connection (i.e. with attracting force) should be located closer to each other. See if repulsive force can be lowered.
- [x] find a better algorithm which does not create islands

### LINKED ISSUE #310: Fix svg group outline glitches in large groups
Outlines around groups many many (unconnected) items are often broken:

image

**Expected behavior**
Pathes do not have unexpected end, no lines drawn across group area. also relates to #307

bonndan:
closes #307 
closes #310 
closes #264 

Also an important step towards the new design (this is the debug view)

image


src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -18,47 +19,86 @@
...
             inArea.add(hex);
             hex.neighbours().forEach(neigh -> {
                 if (!occupied.contains(neigh))
                     inArea.add(neigh);
             });
 
-            //add all "inner" relations (paths)
-            pathsWithinGroup.forEach(rel -> inArea.addAll(rel.getHexes()));
+            Set<Hex> closestNeighbours = getClosestItemsHexes(item, items, allVertexHexes);
+            PathFinder pathFinder = new PathFinder(Set.of()); //we dont care for occupied tiles here
```
mfbieber, line 49: Why don't we care?

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -18,47 +19,86 @@
...
+            closestNeighbours.forEach(neighbour -> {
+                HexPath path = pathFinder.getPath(hex, neighbour);
+                Set<Hex> padded = new HashSet<>(); //pad to avoid thin bridges, also workaround for svh outline issue
```
mfbieber, line 52: Can this be reflected by tests?

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -18,47 +19,86 @@
...
+                    }
+                    if (distance < minDist.get()) {
+                        minDist.set(distance);
+                        min.clear();
+                        min.add(dest);
+                    }
+                });
+
+        return min;
+    }
```
mfbieber, line 100: Is this logic tested by our tests?

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
```diff
@@ -65,6 +68,10 @@ public Hex(int q, int r, int s) {
         this.s = s;
     }
 
+    public Hex(int q, int r) {
+        this(q, r, (r + q) * -1);
+    }
+
```
mfbieber, line 74: Where does this magic formula come from? :)

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
bonndan: I'll improve the comment

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
bonndan: https://www.redblobgames.com/grids/hexagons/implementation.html

TLDR: q and r and like x and y and sufficient to describe a hex position. s is a third axis orthogonal to q and r.

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
bonndan: I've added a test.

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
bonndan: I've added a test.

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
mfbieber: I am trying to really understand this. So I understood that it comes from cube algorithms, and thus s would equal z in a cartesian coordinate system.  

The comment helps to understand that the formula has something to do with a coordinate system, but it does not explain why z = (x + y) * (-1)

In general all these Hex and Layout classes are really hard to understand, without understanding the math and logic behind it. 

Initially we used the linked implementation as a library but then started to implement own logic on top of it, right? 

Having own complex code, without (me) really understanding what it does makes it hard to review and grow it. Isn't there a possibility to use the library as it was and decorate (or maybe another pattern is better suited) with our own logic? That own logic can be well tested and understandable :)

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
```diff
@@ -103,11 +120,19 @@ public static Hex of(long x, long y, float scaling) {
...
+    /**
+     * Returns the distance to the target hex in number of tiles.
+     *
+     * @param target target hex
+     * @return number of tiles
+     */
     public int distance(Hex target) {
-        return lengths(this.subtract(target));
+        Hex hex = this.subtract(target);
+        double l = (Math.abs(hex.q) + Math.abs(hex.r) + Math.abs(hex.s)) / 2.0;
```
mfbieber, line 131: Why is the distance the sum of all coordinates divided by 2?

src/main/java/de/bonndan/nivio/output/map/hex/HexMap.java
```diff
@@ -63,16 +63,7 @@ public Hex hexForItem(Item item) {
...
-
-        optional.ifPresent(hexPath -> {
-            if (start.getGroup() != null && start.getGroup().equals(target.getGroup())) {
-                hexPath.setGroup(start.getGroup());
-            }
-            paths.add(hexPath);
-        });
-
-        return optional;
+        return Optional.ofNullable(pathFinder.getPath(hexForItem(start), hexForItem(target)));
```
mfbieber, line 66: Why doesn't the Pathfinder return an Optional itself?

bonndan: Because nobody refactored it yet 😄

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
bonndan: No, the hex stuff has never been in a library. That was the layouting (force based positioning stuff). Some time ago I investigated some Java hex libraries, but decided against them. If we find a suitable one, we can consider replacing some of our code.

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
bonndan: https://www.redblobgames.com/grids/hexagons/#distances I will add this link as comment, too.

src/main/java/de/bonndan/nivio/output/map/hex/Hex.java
bonndan: Btw, I understood the "s" coordinate as an alternative coordinate, and not as "z". Think of a square with the two sides running parallel to x and y axes in a cartesian system: the sides can be described by offsetting the axes. A hex system has three axes, but you only need to know the offsets of two axes to construct a hex at a position.

src/main/java/de/bonndan/nivio/output/map/hex/HexMap.java
bonndan: I have refactored it to use Optional.

PR #316, opened 2020-10-04
JUnit cleanup

mureinik:
The project uses JUnit Jupiter 5.4.0, but still has the old JUnit 4.12 in its classpath via transitive dependencies. This leads to the fact that code from JUnit 4.12 can be (and is!) used even though it's not supposed to be, and may break if the transitive dependencies upgrade their own dependencies.

This PR cleans up these usages so the test suite uses strictly JUnit Jupiter classes.

It includes the following changes (see individual commits for additional details)
- Removes unused JUnit 4 imports
- Uses assertions from `org.junit.jupiter.api.Assertions` instead of `org.junit.Assert` or `junit.framework.TestCase`
- `IndexerIntegrationTest` used JUnit 4's `FixMethodOrder` annotation to fix the order in which the tests are executed, which did not work, as this annotation is not honored by the JUnit Jupiter runner. Since the test passes and the method order isn't important any more, this annotation was removed.
- `KubernetesTest` explicitly initialized a `KubernetesServer` instance, in JUnit 4 style. The test was rewritten to use the modern `EnableKubernetesMockClient` test extension, and in the process remove some boiler-plate code associated with the `KubernetesServer`.

### DIFF
--- a/src/test/java/de/bonndan/nivio/input/IndexerIntegrationTest.java
+++ b/src/test/java/de/bonndan/nivio/input/IndexerIntegrationTest.java
@@ -4,7 +4,6 @@ import de.bonndan.nivio.input.dto.LandscapeDescription;
 import de.bonndan.nivio.input.dto.ItemDescription;
 import de.bonndan.nivio.model.*;
 import de.bonndan.nivio.output.LocalServer;
-import org.junit.Assert;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -229,7 +228,7 @@ public class IndexerIntegrationTest {
         LandscapeImpl landscape = index("/src/test/resources/example/example_templates.yml");
 
         LandscapeItem web = landscape.getItems().pick("web", null);
-        Assert.assertNotNull(web);
+        assertNotNull(web);
         assertEquals("web", web.getIdentifier());
         assertEquals("webservice", web.getType());
     }
--- a/src/test/java/de/bonndan/nivio/input/IndexerIntegrationTest.java
+++ b/src/test/java/de/bonndan/nivio/input/IndexerIntegrationTest.java
@@ -5,11 +5,9 @@ import de.bonndan.nivio.input.dto.ItemDescription;
 import de.bonndan.nivio.model.*;
 import de.bonndan.nivio.output.LocalServer;
 import org.junit.Assert;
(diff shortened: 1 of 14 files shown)


bonndan:
Now that's an awesome PR! CI failed, but due to an maven download connection failure. The tests have been restarted and I'll wait for the result.

I've also added the Hacktoberfest label in case it is needed.

src/test/java/de/bonndan/nivio/input/IndexerIntegrationTest.java
```diff
@@ -30,7 +29,7 @@
 @SpringBootTest
 @ExtendWith(MockitoExtension.class)
 @ActiveProfiles("test")
-@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+@TestMethodOrder(Alphanumeric.class)
```
bonndan, line 32: Thanks for your efforts, but method ordering is not necessary anymore. I've checked that the tests run without it. It also does not make sense anymore because nivio works differently now since when method ordering was introduced.

I think you can remove it.

src/test/java/de/bonndan/nivio/input/KubernetesTest.java
```diff
@@ -18,22 +17,16 @@
 import java.util.List;
 import java.util.Map;
 
-import static org.junit.Assert.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
+@EnableKubernetesMockClient(crud = true)
```
bonndan, line 23: 👍

bonndan (commented):
No complaints. Could you try running IndexerIntegrationTest without fixed method order?

src/test/java/de/bonndan/nivio/input/IndexerIntegrationTest.java
mureinik: Good, point, will do.

mureinik: Done.
Leaving the conversation unresolved for your review.

mureinik:
@bonndan I removed the test ordering on `IndexerIntegrationTest` as per your comment and pushed a new patchset. This in turn retriggered the CI, which now passes (as you commented - the failure was unrelated to the PR itself, and was just maven failing to pull some dependency).

PR is now ready for your review.

bonndan:
Thanks for your contribution!

PR #320, opened 2020-10-12
new icon set

### LINKED ISSUE #262: Make map icons independent of local web server
- [ ] we could embed base64-encoded images
- [ ] or embed svg fonts (https://graphicdesign.stackexchange.com/questions/5162/how-do-i-embed-google-web-fonts-into-an-svg)

relates to #314

### LINKED ISSUE #300: Enhance documentation how to modify map apperance
- [x] add an example preview in the docs of  item background fill
- [x] document how to style a map with external css (see dedica.yml as example)

### LINKED ISSUE #314: Find a new icon set
Nivio needs a new icon set reflecting the many possible landscape item types (all kinds of network, software, architecture stuff like server, api, firewall, you name it).

Requirements:

- [x] has a clear and commercially compatible license (preferably MIT) and can be shipped with nivio
- [x] contains SVG icons
- [x] modern appearance (material ui, slim, lean and similar)
- [ ] must NOT contain any icons showing protected brands or trademarks

bonndan:
* closes #314 
* closes #300 
* closes #262 

- [x] downloads icons via npm and then copies the icons to java server resource. We still need to ensure that this works everywhere (currently "cp" is used) and that the order or building is correct and icons land in the final jar.


src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
```diff
@@ -0,0 +1,115 @@
...
+        String iconName = IconMapping.of(type.toLowerCase()).map(IconMapping::getIcon).orElseGet(type::toLowerCase);
+        return getIconUrl(iconName, true);
+    }
+
+    /**
+     * Provides an URL for a locally served icon.
+     *
+     * @return an url pointing to a file or a data url
+     */
+    String getIconUrl(String icon, boolean fallback) {
```
mfbieber, line 80: Can't this be private?

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public String getIconUrl(Item item) {
+
+        //icon label based
+        String icon = item.getIcon();
+        if (!StringUtils.isEmpty(icon)) {
+
+            if (icon.startsWith(VendorIcons.VENDOR_PREFIX)) {
+                String key = icon.replace(VendorIcons.VENDOR_PREFIX, "").toLowerCase();
+                return vendorIcons.getUrl(key).orElse(defaultIcon);
```
mfbieber, line 55: This is the only place where `vendorIcons.getUrl()` is used, so I was at first a little confused about what `LocalIcons` really is. I thought `LocalIcons` are icons addes by the user? Is this correct? (In this case, maybe a different naming would be better. Why not just Icons?)

Or does it mean that there are not retrieved from a remote location such as the vendor icons?

src/main/app/copy-icons.js
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+const fs = require('fs-extra');
+
+// Async with promises: 
+fs.copy('node_modules/@mdi/svg/svg', '../resources/static/icons/svg')
+    .then(() => console.log('Icons copied successfully!'))
+    .catch(err => console.error(err));
```
mfbieber, line 7: This is what you mean with "cp"?

I checked the build with Windows 10, it works fine. I will now check with MacOS.

docs/source/output.rst
```diff
@@ -20,31 +20,48 @@ Nivio indexes all landscape items in an in-memory search engine called Lucene. Y
...
-Vendor Icons
-------------
+.. code-block:: yaml
+   :linenos:
 
+    items:
+      - identifier: bar
+        type: account
+
+The following types are translated icon to maintain backward compatibility:
```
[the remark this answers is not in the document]
bonndan: true...

docs/source/output.rst
```diff
@@ -56,6 +73,52 @@ Vendor icons are work in progress.
...
+.. code-block:: yaml
+   :linenos:
+
+   items:
+      - identifier: bar
+        shortname: FOOBAR
+      - identifier: pony
+        shortname: 🦄
+
+
```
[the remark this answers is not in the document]
bonndan: I knew you'd like the 🦄

src/main/app/copy-icons.js
bonndan: "cp" is not used anymore. This is copied production code.

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: I've been lazy and made it package-private for testing. 🙈

src/main/app/copy-icons.js
mfbieber: Fine, I did not find "cp", I am checking with MacOS anyways :)

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: LocalIcons is a crappy name, but I had no better idea. It stands for the locally (fs) available svg icons that come shipped with nivio. They are loaded and embedded, so nivio does not serve anything anymore.

mfbieber: Okay, I think for the users it is okay. As I developer, I now understand. Maybe we can add a brief comment to the class, so others developers understand too?

In this case I also did not come up with a better name. "BuiltInIcons" could be another idea.

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
mfbieber: But I did not see it being used in the tests - did I overlook this?

mfbieber:
The JAR contains the icons both in Windows and MacOS as desired!

Good Job @bonndan 😻

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: LocalIconsTest uses it. Maybe it were better to return an Optional and get rid of the boolean flag.

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: One more ugly aspect is that it is named to handle local icons, but suddenly comes up with remote vendor icons. I don't like my design here.

mfbieber: Yes, that is exactly what I was wondering about.

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
mfbieber: Oh, yes.

Yes, that would be better!

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: refactored.

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: done

src/main/java/de/bonndan/nivio/output/icons/IconService.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+
+
+/**
+ * Provides the builtin icons (shipped with nivio) and vendor icons (loaded form remote locations) as embeddable data.
+ *
+ *
+ */
+@Service
+public class IconService {
```
mfbieber, line 18: This is a lot better 👍

PR #327, opened 2020-10-18
Rendering optimizations

### LINKED ISSUE #325: Reduce SVG map file size
A map of about 90 services and several groups and connections results in a SVG file of 11MB. 

** Proposed solution **
It might be that the generation of background hexes and the embedding of images as data-urls can be optimised

- [x] can svgs use repeated background images?
- [ ] see if embedded icons (data-url) can be reused as patterns
- [ ] check if using utf-8 symbol ⚙ can be used as default (looks different than icon set, so not best solution)

### LINKED ISSUE #326: Items not contained within group boundaries
In the large graph test data the following occurs:

image

bonndan:
closes #325 (demo map svg reduced from 680kb to 180kb)
closes #326 strategy to build group area is to connect each to closest neighbour, but prevent reverse connections


bonndan:
Island-like group outlines are created

image

We could easily smoothen this to reactange shaped outlines, but I am not sure is this might raise collision conflicts with other areas and the areas start to look a bit clunky

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -34,37 +35,57 @@
...
 
-        Set<Hex> bridges = getBridges(inArea);
-        inArea.addAll(bridges);
+        Set<Hex> bridges = getBridges(inArea, 2);
+        while (!bridges.isEmpty()) {
+            inArea.addAll(bridges);
+            bridges = getBridges(inArea, 3); // 2 might be too aggressive and collide with other group areas
+        }
 
         return inArea;
```
mfbieber, line 90: This logic is hard to review. Maybe it is best if we talk this through in a call? 

Are our tests good enough? I saw that you did not need to make many changes there :)

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -75,41 +96,42 @@
...
      * @param allVertexHexes item hex mapping
+     * @param connected
      * @return the closest neighbours
      */
-    private static Set<Hex> getClosestItemsHexes(Item item, Set<Item> items, Map<Item, Hex> allVertexHexes) {
+    private static Optional<Item> getClosestItem(Item item, Set<Item> items, Map<Item, Hex> allVertexHexes, List<Item> connected) {
         Hex start = allVertexHexes.get(item);
         AtomicInteger minDist = new AtomicInteger(Integer.MAX_VALUE);
-        final Set<Hex> min = new HashSet<>();
+        AtomicReference<Item> min = new AtomicReference<>(null);
```
mfbieber, line 105: Where does the necesity of the AtomicReference arise? What could (concurrently) change/read that reference?

src/main/java/de/bonndan/nivio/output/map/svg/SVGBackgroundFactory.java
```diff
@@ -0,0 +1,42 @@
...
+                ContainerTag use = SvgTagCreator.use("#" + HEX)
+                        .attr("x", (int) hex.x - 2 * Hex.HEX_SIZE)
+                        .attr("y", y);
+                background.add(use);
+            }
+            i++;
+        }
+
+        return background;
+    }
```
mfbieber, line 41: This is also tough to review, we could discuss this in a call. 

I did not find a test for this 😨, we should have one!

src/main/java/de/bonndan/nivio/output/map/svg/SVGDocument.java
```diff
@@ -120,16 +123,14 @@ public DomContent render() {
...
-        var i = 0;
-        for (int q = minQ.get(); q <= maxQ.get(); q++) {
-            for (int r = minR.get() - i; r <= (maxR.get() + (maxQ.get())); r++) {
-                background.add(new SVGHex(new Hex(q, r), "none", "#cccccc").render());
-            }
-            i++;
-        }
-
-        int paddingTopLeft = 2 * Hex.HEX_SIZE;
-
```
mfbieber, line 132: Is this the logic we now have in `SVGBackgroundFactory`?

src/main/java/de/bonndan/nivio/output/map/svg/SVGItemLabel.java
```diff
@@ -2,60 +2,39 @@
...
-        if (!StringUtils.isEmpty(item.getOwner()))
-            g.attr("data-owner", item.getOwner());
-
-        Map<String, String> groupedLabels = Labeled.groupedByPrefixes(item.getLabels());
-        groupedLabels.forEach((key, value) -> {
-            if (!StringUtils.isEmpty(value)) {
-                g.attr("data-" + key, value);
-            }
-        });
-
```
mfbieber, line 54: Why don't we need this anymore?

src/main/java/de/bonndan/nivio/model/ServiceInterface.java
```diff
@@ -18,7 +18,6 @@
     public ServiceInterface() {
     }
 
-    //TODO: who uses these constructors?
     public ServiceInterface(InterfaceDescription interfaceItem) {
```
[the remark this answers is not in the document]
bonndan: The other way round: the first one is unused. I have removed it now.

src/main/java/de/bonndan/nivio/output/Renderer.java
```diff
@@ -12,7 +12,7 @@
  */
 public interface Renderer<T> {
 
-    T render(LayoutedComponent landscape);
+    T render(LayoutedComponent layoutedComponent, boolean debug);
 
-    void render(LayoutedComponent landscape, File file) throws IOException;
+    void render(LayoutedComponent landscape, File file, boolean debug) throws IOException;
```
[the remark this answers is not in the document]
bonndan: I've added some documentation to the interface which hopefully explains my intention. Although the second method is not used yet, I'd like to keep it. There have been different renderers before (e.g. png binary output), and maybe there will be others (3D stuff?).

src/main/java/de/bonndan/nivio/output/map/MapController.java
```diff
@@ -33,14 +30,16 @@ public MapController(LandscapeRepository landscapeRepository, RenderCache render
...
+                                      @RequestParam(value = "debug", required = false, defaultValue = "false") boolean debug
+    ) {
         Landscape landscape = getLandscape(landscapeIdentifier);
 
         try {
             HttpHeaders headers = new HttpHeaders();
             headers.add(HttpHeaders.CONTENT_TYPE, "image/svg+xml");
             return new ResponseEntity<>(
-                    renderCache.getSVG(landscape),
+                    renderCache.getSVG(landscape, debug),
```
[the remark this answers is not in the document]
bonndan: Yes, the svg map endpoint accepts a request param. However, I'd like to keep it hidden for now.

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
bonndan: It's the stream below which requires a (effectively) final variable.

src/main/java/de/bonndan/nivio/output/map/svg/SVGDocument.java
bonndan: Yes

src/main/java/de/bonndan/nivio/output/map/svg/SVGItemLabel.java
bonndan: My initial idea was to stuff as much data into the svg as possible, so that data can be retrieved by any post-processors without the need to call nivio again. This is required yet and comes at the cost of large file sizes, so I think it's better to remove it for now.

bonndan:
I couldn't sleep, so I luckily figured out how to improve pathfinding, so that group areas are omitted if possible. This results in much nicer connections to be drawn.

image

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -26,95 +28,112 @@
...
      *
-     * @param occupied       tiles occupied by items
-     * @param group          the group
-     * @param allVertexHexes a mapping from item to its hex (all, unfiltered)
+     * @param hexesToItems tiles occupied by items
+     * @param group        the group
      * @return all hexes the group consists of (an area)
      */
-    public static Set<Hex> getGroup(Set<Hex> occupied, Group group, Map<Item, Hex> allVertexHexes) {
+    public static Set<Hex> getGroup(BidiMap<Hex, Object> hexesToItems, Group group) {
```
mfbieber, line 35: Which special functionality provided by the BidiMap data structure are we using? I have only seen `getKey()` and `containsKey()`. If that is so, why isn't that just the value in a normal Map structure?

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
```diff
@@ -26,95 +28,112 @@
...
-        List<Item> connected = new ArrayList<>();
 
         if (!items.iterator().hasNext()) {
             LOGGER.warn("Could not determine group area for group {}", group);
             return inArea;
         }
-        //surround each item
+
+        //build the area by adding paths
+        addPathsBetweenClosestItems(hexesToItems, items, inArea);
```
mfbieber, line 46: Great! Extracting this into an own method makes everything better readable and understandable 😄

src/main/java/de/bonndan/nivio/output/map/svg/SVGBackgroundFactory.java
```diff
@@ -9,6 +9,11 @@
 
 import static java.lang.Math.round;
 
+/**
+ * This generates the background hex tiles.
+ *
+ * TODO It is a bad bank for technical debt.
```
mfbieber, line 15: What technical debt do you think that we are collecting here? For me it is just that I don't really understand what is going on (but I also haven't fully understood the other rendering logics).

src/test/java/de/bonndan/nivio/output/map/hex/GroupAreaFactoryTest.java
```diff
@@ -2,12 +2,15 @@
...
 import org.apache.commons.collections4.set.UnmodifiableSet;
 import org.junit.jupiter.api.Test;
 
 import java.util.*;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.in;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 class GroupAreaFactoryTest {
```
mfbieber, line 16: Do these tests cover the logic of finding the closest items and the correct occupation of tiles as implemented in the methods `getGroup()`, `addPathsBetweenClosestItems()`, `getClosestItem()` and `getBridges()`?

src/main/java/de/bonndan/nivio/output/map/svg/SVGBackgroundFactory.java
bonndan: Here we produce hundreds of hexagons to display them in the background. There is no real added value and no business logic, but the designer suggested it like that.

src/test/java/de/bonndan/nivio/output/map/hex/GroupAreaFactoryTest.java
bonndan: Yes, but it's more like a blackbox test, i.e. we do not test the exact behaviour, but the outcome.

src/main/java/de/bonndan/nivio/output/map/svg/SVGBackgroundFactory.java
bonndan: I have refactored it a bit and added some comments.

src/main/java/de/bonndan/nivio/output/map/hex/GroupAreaFactory.java
bonndan: Excellent observation. It has been refactored.

src/test/java/de/bonndan/nivio/output/map/hex/GroupAreaFactoryTest.java
mfbieber: Okay, is it hard to add tests for the steps in between? The problem to me is, that it is hard to understand what is going on and also hard to refactor if we only have blackbox tests :-/

src/main/java/de/bonndan/nivio/output/map/svg/SVGBackgroundFactory.java
mfbieber: Nice, it is better readble now!

Issue #97, opened 2020-01-30
Link code analysis artifacts


bonndan:
"As an architect I want to drill down on landscape item details like code analysis artifacts".

Provide a means to assign links more easily than inserting them manually (one way is label parsing). E.g. configure a code analysis system like Sonarcube globally and map items to certain projects automatically. 

related to #92

bonndan:
What about this approach, namely having links with special keys:

```yaml
items:
   - identifier: blog-server
     links:
       sonarqube: https://sonarcloud.io/dashboard?id=foo
```
The "sonarqube" key could be handled by some specialized SonarQube KPI resolver .
However, this would require a lot of copy/paste work and would hardly work with authentication. The approach mentioned in #92 would support authentication, since a preconfigured KPI resolver work work in a similar way as the resolvers fetching files from protected git repos.

bonndan:
@sl0815 I've added a draft how this could be wired and how it could work. I'm not saying it must be implemented this way, it's only to introduce some of the internal parts of nivio. 

The root of all evil can be found in Indexer.java.

sl0815:
@bonndan Thanks a lot. I'll have a look at it.

bonndan:
@Matthimatiker this is a part of the feature we talked about.

PR #329, opened 2020-10-31
working towards more immutable Landscape object

mfbieber:


src/test/java/de/bonndan/nivio/output/map/RenderCacheTest.java
```diff
@@ -43,41 +43,48 @@ public void setup() {
...
+        );
+        Landscape two = getLandscape("second", "two");
         two.setProcessLog(one.getLog()); //sync last update
-        two.setIdentifier("second");
         String second = renderCache.getSVG(two, false);
 
-        verify(stylesheetFactory, times(2)).getMapStylesheet(any(), any());
+        verify(stylesheetFactory, times(2))
+                .getMapStylesheet(any(), any());
     }
```
mfbieber, line 67: If we work towards immutable objects, the identifier must not be set from the outside. Did I understand the Test as it was before and do you think that it is meaningful now too?

bonndan: You're right. It makes no sense changing the identifier.

bonndan: The test is still valid.

mfbieber: Thanks :)

src/main/java/de/bonndan/nivio/api/ApiController.java
```diff
@@ -133,8 +133,7 @@ public ProcessLog indexLandscape(
             @RequestHeader(name = "format") String format,
             @RequestBody String body
     ) {
-        LandscapeDescription dto = new LandscapeDescription();
-        dto.setIdentifier(identifier);
+        LandscapeDescription dto = new LandscapeDescription(identifier, "", null);
```
mfbieber, line 136: Is this meaningful?

This is for `@RequestMapping(path = "/landscape/{identifier}/services", method = RequestMethod.POST)`, which I think triggers the indexing of a landscape provided by the POST body? Is this correct? Or are just Items added?

mfbieber:
I have some questions regarding the fields in `Landscape`:
```java
    /**
     * Immutable unique identifier. Maybe use an URN.
     */
    @NonNull
    @Pattern(regexp = IDENTIFIER_VALIDATION)
    private final String identifier;

    /**
     * Human readable name.
     */
    @NonNull
    private final String name;

    /**
     * Maintainer email
     */
    @Nullable
    private String contact;

    private String description;

    private String source;

    @JsonIgnore
    private final ItemIndex items = new ItemIndex();

    private LandscapeConfig config;

    private final Map<String, Group> groups = new HashMap<>();

    private ProcessLog processLog;

    private final Map<String, String> labels = new HashMap<>();
    private final Map<String, Link> links = new HashMap<>();
    private String owner;
```

I saw that by the yml files, `contact` can be null. How about `description`, `source`, `config` and `owner`?

And I guess `items`, `groups`, `labels` and `links` must be set after Object creation, correct? To make `Landscape` completely immutable, we would need to refactor the way these fields are retrieved and then at the end create the `Landscape` object. Do you think this is doable?

And what about the `ProcessLog`, is this also something updated after the `Landscape` object has been created? Is this updated whenever `items`, `groups`, `labels` and `links` are added?

src/main/java/de/bonndan/nivio/api/ApiController.java
bonndan: This is for adding items via POST body. It should also trigger reindexing. While you're at it, you might consider renaming the path to "/items", because "/services" is not used as term anymore and narrows the usage too much (although this is not scope of the ticket).

bonndan:
> And I guess `items`, `groups`, `labels` and `links` must be set after Object creation, correct? To make `Landscape` completely immutable, we would need to refactor the way these fields are retrieved and then at the end create the `Landscape` object. Do you think this is doable?

For sure it's feasible, but I'd ask myself if it is worth the big effort. I'd rather not refactor it just for sake of having "better software", but no added value. Whenever mutability leads to unclear situations, we should remove it, of course. Maybe having a "locking mechanism" leads to the same result.

In addition, when #97 is implemented, we'll have a lot of async data coming in.

> And what about the `ProcessLog`, is this also something updated after the `Landscape` object has been created? Is this updated whenever `items`, `groups`, `labels` and `links` are added?

Yes, and other situations, too. ProcessLog is meant to be extended all the time.

src/main/java/de/bonndan/nivio/api/ApiController.java
mfbieber: That is fine, I renamed the endpoint.

So it is totally ok to have a LandscapeDescription with an identifier, but without a human readble name - correct?

src/main/java/de/bonndan/nivio/assessment/StatusValue.java
```diff
@@ -54,23 +63,26 @@
...
+        }
+        return new StatusValue(fieldName, maxValue.getStatus(), maxValue.getMessage(), maxValue.getField());
     }
 
     public StatusValue(@NonNull String field, @Nullable Status status, @Nullable String message) {
         if (StringUtils.isEmpty(field)) {
             throw new IllegalArgumentException("Status value has no field");
         }
-        this.field = field;
+        this.field = Objects.requireNonNull(field);
```
bonndan, line 85: StringUtils.isEmpty also handles null. This is why I like it.

mfbieber: Yes, I know that and use it too for null and empty checks. I don't know why I did that 😄

mfbieber:
> For sure it's feasible, but I'd ask myself if it is worth the big effort. I'd rather not refactor it just for sake of having "better software", but no added value. Whenever mutability leads to unclear situations, we should remove it, of course. Maybe having a "locking mechanism" leads to the same result.

@bonndan: well, we could argue about value and if these kind of refactorings lead to a better mantainability. In my opinion there is a lot of value in that. But I wouldn't for sure want to overengineer the whole thing if it does not really help us.

I would concentrate on the most crucial fields though. Which fields are absolutely necessary to have a healthy `Landycape`? These could be the ones needed for inmutability.

bonndan:
> I would concentrate on the most crucial fields though. Which fields are absolutely necessary to have a healthy `Landscape`? These could be the ones needed for inmutability.

Exactly. I totally agree. The input DTOs are just short-lived objects for single use. We'll just shoot ourselves in the foot if the throw all the fine patterns at them.

src/main/java/de/bonndan/nivio/api/ApiController.java
bonndan: Yes.

mfbieber:
Could you help me identify which of the fields  are crucial for `Landscape` - from a domain design point of view?

And could you also say something about the nullability of the fields too?
> I saw that by the yml files, contact can be null. How about description, source, config and owner?

In addition, is this correct, or is it possible to set any of these upon object creation?
> And I guess items, groups, labels and links must be set after Object creation, correct? To make Landscape completely immutable, we would need to refactor the way these fields are retrieved and then at the end create the Landscape object. Do you think this is doable?

bonndan:
* description might be null
* owner might be null
* source NOT NULL
* config NOT NULL (although an "empty" config is possible)
* label / links collections can be immutable and final
* even the items collection (ItemIndex) could be final if we change how DiffProcessor works

How will updates be handled in the future?

We must be aware of "partial" updates, i.e. when the sources required to build the input DTOs cannot be resolved completely, we apply what is available, but do not delete any items or relations.

mfbieber:
Thanks!

> How will updates be handled in the future?

> We must be aware of "partial" updates, i.e. when the sources required to build the input DTOs cannot be resolved completely, we apply what is available, but do not delete any items or relations.

Well, if we need to update any information that is crucial for the object creation (such as the identifier), we should create a new object.

If that object has relations with other objects (or items belonging to it), these could be added to the new object as well. 

But if we construct a new object, do we really want to keep the old references to items and relations? Is it that hard to resolve them once again? From a clean point of view, this approach would be better.

bonndan:
If we work on new objects and have incomplete input data only, we can only resolve what is present and will lose existing relations. However, this is a question of point of view. The current solution of doing partial updates and no deletes has flaws, too.

Issue #321, opened 2020-10-15
Refactoring KPI initialization to run earlier


bonndan:
When the KPIFactory inits the KPIs, the related landscape has long been parsed, although KPI config is just a part of the common config (a field in LandscapeDescription). As a result, KPI config errors are raised way too late (when assessment is run) and not when a user submits the landscape input data.

Fix:

Initialize KPIs "immediately" when Indexer runs.

relates to #319

PR #331, opened 2020-11-24
[#321] kpis initialized early with landscape

bonndan:


src/main/java/de/bonndan/nivio/assessment/AssessmentController.java
```diff
@@ -31,9 +28,7 @@ public AssessmentController(LandscapeRepository landscapeRepository, KPIFactory
             return ResponseEntity.notFound().build();
         }
 
-        Assessment assessment = new Assessment(
-                landscape.applyKPIs(factory.getConfiguredKPIs(landscape))
-        );
+        Assessment assessment = new Assessment(landscape.applyKPIs(landscape.getKpis()));
```
mfbieber, line 31: This looks a bit strange at the first sight. I checked more deeply and realized that `applyKPIs()` is a default method in the `Asseable` interface that recursively calls itself. 

I would have suggested to make `applyKPIs()` in `Landscape` use the KPIs already known to the `Landscape`. I did not see any usage of `landscape.applyKPIs(Map<String, KPI> kpis)` where we would provide the KPIs from the outside.

mfbieber: I think we could check how the `Landscape` object works after https://github.com/dedica-team/nivio/issues/212 has been resolved. Maybe thinks like this are already cleaned up after these changes.

src/main/java/de/bonndan/nivio/assessment/kpi/KPIFactory.java
```diff
@@ -32,16 +29,18 @@ public KPIFactory(ApplicationEventPublisher eventPublisher) {
...
+     * @param kpiConfigMap the landscape kpi configurations
      * @return effective KPIs
+     * @throws ProcessingException if a KPI is misconfigured
      */
-    public Map<String, KPI> getConfiguredKPIs(Landscape landscape) {
+    public Map<String, KPI> getConfiguredKPIs(@NonNull Map<String, KPIConfig> kpiConfigMap) {
 
-        Map<String, KPIConfig> config = landscape.getConfig().getKPIs();
-        Map<String, KPI> kpis = new HashMap<>(defaultKPIs.size() + config.size());
+        Objects.requireNonNull(kpiConfigMap, "kpi config is null or empty");
```
mfbieber, line 38: `Objects.requireNonNull()` does not check if the map is empty, only if an object == null.

mfbieber:
While reviewing I came across `reIndex(final LandscapeDescription input)` in `Indexer`. I tried to figure out if we also have the `index()` case with a `LandscapeDescription`, but did not find it.

Does that case exist?

And do the oder `index()` methods we have basically do the same thing as the `reIndex()`? Or asked differently: is "index" the same thing for all these methods?

bonndan:
We had no "index" method until now. If chose the name to indicate that the method would also update existing landscape, but now there is a comment and the method is named "index". 👍

src/main/java/de/bonndan/nivio/assessment/kpi/KPIFactory.java
bonndan: Yes. If the are no configured KPIs, the map would be empty.

mfbieber: I meant that the message is missleading, we are not checking if the map is empty, only if it is null ;)

bonndan: Fixed in a different branch.

Issue #332, opened 2020-11-27
Laggy navigation on map


bonndan:
**Describe the bug**
A *large* map is terrible to scroll/drag using the user inferface (map view).

**To Reproduce**
Load the gui, enter a map. The view starts in the upper left corner, which is usually empty except for the title.

**Expected behavior**
* smooth scrolling
* zoomed out view OR zoomed on center OR small window to contain an overview (currently disabled)

**Screenshots**
Customer data, cannot be shown here.

** Possible solution **

* remove the repeated background tiles, which have to be rendered in hundreds or more

- [x] the issue is fixed!
- [x] fix duplicate loading of map svg

Issue #303, opened 2020-09-22
Proposal how to display landscape item assessment result (status) in map 


bonndan:
At the moment the map does not show any assessment (status) information. This information can be retrieved by clicking items in the browser based map navigator, but the bare printable SVG does not contain this data.

- [ ] make a proposal how to display status information near the items

bonndan:
Code from SVGItem showing "scale".

`
    /* TODO make scale reappear, but without evaluation (this is part of assessment)
    private ContainerTag getScale() {

        if (StringUtils.isEmpty(item.getLabel(Label.SCALE))) {
            return null;
        }

        int scaleVal = 0;
        try {
            scaleVal = Integer.parseInt(item.getLabel(Label.SCALE));
        } catch (NumberFormatException ignored) {
        }

        return SvgTagCreator.g(
                        SvgTagCreator.circle()
                                .attr("cx", 0)
                                .attr("cy", 0)
                                .attr("r", 12)
                                .attr("fill", scaleVal > 0 ? "green" : "red")
                        ,
                        SvgTagCreator.text(String.valueOf(scaleVal))
                                .attr("transform", "translate(-" + 4 + "," + 5 + ")")
                ).attr("transform", "translate(" + 30 + "," + 30 + ")");
    }

     */
`

bonndan:
One possibility is to add small badges to the item circles:

![image](https://user-images.githubusercontent.com/143695/104363818-06ffc900-5516-11eb-8e13-2752ad835a3f.png)
Screenshot taken from https://material-ui.com/components/avatars/#avatar

bonndan:
![image](https://user-images.githubusercontent.com/143695/111473115-fb559c00-872a-11eb-8215-b77fef80ac9e.png)

bonndan:
Background blur approach is used for now.

PR #334, opened 2020-11-29
332 fixes

### LINKED ISSUE #304: Make map items (circles) active/clickable
At the moment ony the name below map items are clickable to make item information appear. If the user clicks an icon/circle, nothing happens, although this might be the expected behaviour.

- [ ] copy behaviour of the item label to the item circle, so that information appear on item click, too

bonndan:
* Fixes that SVG is loaded twice (cleanup of previous #332 fix attempt)
* closes #304 (small UX improvement)


src/main/java/de/bonndan/nivio/output/map/svg/SVGItem.java
```diff
@@ -81,11 +81,11 @@ public DomContent render() {
             circle.attr("stroke-dasharray", 5);
             circle.attr("opacity", 0.7);
         }
-        ContainerTag inner = SvgTagCreator.g(circle, content, children)
-                .attr("class", "hexagon");
+        ContainerTag inner = SvgTagCreator.g(circle, content, children);
 
         return SvgTagCreator.g(inner, icon)
-                .attr("class", "hexagon-group")
```
mfbieber, line 88: Was this what made the hexagons before? Are we not having the hexagons anymore?

mfbieber: Starting in line 92 in this class there is a TODO and a commented private method. Is this dead code? Do we need it, or can we remove it?

src/main/java/de/bonndan/nivio/output/map/svg/SVGItemLabel.java
```diff
@@ -27,11 +27,10 @@ public DomContent render() {
...
 
-        ContainerTag g = SvgTagCreator.g(null, labelText).attr("class", "label");
-        g.attr("id", id);
-        g.attr("data-identifier", identifier);
-        g.attr("text-anchor", "middle");
-        return g;
+        return SvgTagCreator.g(null, labelText)
+                .attr("class", "label")
+                .attr("id", id)
+                .attr("text-anchor", "middle");
```
mfbieber, line 33: What does removing the "data-identifier" attribute do?

src/main/java/de/bonndan/nivio/output/map/svg/SVGItem.java
```diff
@@ -81,11 +81,11 @@ public DomContent render() {
...
+                .attr("data-identifier", this.id)
+                .attr("class", "item")
```
mfbieber, line 88: What does the data-identifier and the class attribute change here?

src/test/java/de/bonndan/nivio/output/map/svg/SVGItemLabelTest.java
```diff
@@ -21,8 +24,8 @@ public void regression184() {
         foo.setLandscape(landscape);
 
 
-        SVGItemLabel svgItemLabel = new SVGItemLabel(foo);
-        String render = svgItemLabel.render().render();
+        SVGItem svgItem = new SVGItem(null, new LayoutedComponent(foo), new Point2D.Double(0,0));
+        String render = svgItem.render().render();
```
mfbieber, line 28: How are we still testing the `SVGItemLabel` now?

src/main/java/de/bonndan/nivio/output/map/svg/SVGItemLabel.java
bonndan: "data-identifier" is used in javascript to obtain the item identifier. This has been moved to a parent element (see below).

src/main/java/de/bonndan/nivio/output/map/svg/SVGItem.java
bonndan: The element gets on onClick event (all ".item" elements) and the identifier is used to show correct item data in a dialogue.

src/main/java/de/bonndan/nivio/output/map/svg/SVGItem.java
bonndan: The item is neither a "hexagon" anymore nor do we need the class here anymore.

The dead code has been copied to https://github.com/dedica-team/nivio/issues/303

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
```diff
@@ -3,33 +3,53 @@
...
+     */
+    public LocalIcons(@Value("${nivio.iconFolder:#{null}}") String iconFolder) {
+        if (iconFolder != null) {
+            this.iconFolder = iconFolder.endsWith(File.separator) ? iconFolder : iconFolder + File.separator;
+        } else {
+            this.iconFolder = DEFAULT_ICONS_FOLDER;
+        }
+        defaultIcon = getIconUrl(DEFAULT_ICON.getIcon()).orElseThrow(() -> {
+            throw new RuntimeException(String.format(initErrorMsg, this.iconFolder));
+        });
```
mfbieber, line 52: Okay, the problem I had was indeed resolved by obtaining the npm dependency. 👍

src/main/resources/application.yml
```diff
@@ -28,6 +28,7 @@ nivio:
   baseUrl: ${NIVIO_BASE_URL:}
   version: #project.version#
   pollingMilliseconds: 30000 # 30 secs
+  # iconFolder: /a/local/path
```
mfbieber, line 31: What do we need this for?

mfbieber: Ah, okay, to provide it, if it is different than the default folder 💡

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
mfbieber: Why not making the parameter `@Nonnull` and using the `DEFAULT_ICONS_FOLDER` as the fallback value?

src/main/java/de/bonndan/nivio/output/map/svg/SVGItem.java
mfbieber: Okay, I haven't looked into the frontend logic querying for these yet.

src/main/java/de/bonndan/nivio/output/icons/LocalIcons.java
bonndan: I've had to look up how that is done using injected values, but it works. Change is pushed.

src/test/java/de/bonndan/nivio/output/layout/OrganicLayouterTest.java
```diff
@@ -168,7 +168,7 @@ public void renderLandscapeItemModelWithMagicLabels() throws IOException {
     public void renderCSV() throws IOException {
 
         HttpService httpService = new HttpService();
-        IconService iconService = new IconService(new LocalIcons(null), new VendorIcons(httpService));
```
mfbieber: This looks pretty now 🦋

PR #336, opened 2020-12-07
External data from links, interal refactoring

bonndan:
External data can be loaded using links. Currently it reads GitHub projects and Spring Boot health checks. A LOT can be done here, but I think this is better handled in separate issues.

This escalated quickly. I've refactored and cleaned up some stuff in the Indexer, now there are Resolvers which work on the incoming data, and Processors which modify the landscape.


src/main/java/de/bonndan/nivio/input/linked/SonarLinkHandler.java
```diff
@@ -0,0 +1,41 @@
...
+     * @return
+     */
+    @Override
+    public CompletableFuture<String> resolveAndApplyData(Link link, Labeled component) {
+        if (sonarClient.isEmpty()) {
+            return CompletableFuture.completedFuture("SonarClient is not configured");
+        }
+
+        SonarClient sonarClient = this.sonarClient.get();
+        // TODO
```
bonndan, line 38: @sl0815 still no time to work on this? The GitHub connection is already working. Loading Sonarqube stuff is just a matter of knowing their API.

docs/source/input.rst
```diff
@@ -46,4 +46,24 @@ from environment variables. To exclude internal stacks (like those responsible f
 Nivio proprietary format
 ------------------------
 
-Nivio provided an own format, which allows to set all model properties manually (see Model and Syntax section)
\ No newline at end of file
+Nivio provided an own format, which allows to set all model properties manually (see Model and Syntax section).
```
mfbieber, line 49: This sould be "provides" - it still does, does it?

docs/source/input.rst
```diff
@@ -46,4 +46,24 @@ from environment variables. To exclude internal stacks (like those responsible f
...
+
+
+External data
+-------------
+
+Nivio can load external data that can not be used directly to build landscapes, but is still valuable. For example, the
+number of GitHub issues might be interesting to see on a landscape item that is an open source component. To attach such
+data to landscape components, use links having special known identifiers like "github" or "sonar".
+
```
mfbieber, line 58: It must be "cannot"

src/main/java/de/bonndan/nivio/input/ComponentDescriptionValues.java
```diff
@@ -0,0 +1,60 @@
...
+
+        assignSafeIfAbsent(increment.getName(), item.getName(), item::setName);
+        assignSafeIfAbsent(increment.getDescription(), item.getDescription(), item::setDescription);
+        assignSafeIfAbsent(increment.getOwner(), item.getOwner(), item::setOwner);
+
+        Labeled.merge(increment, item);
+
+        increment.getLinks().entrySet().stream()
+                .filter(entry -> !item.getLinks().containsKey(entry.getKey()))
+                .forEach(entry -> item.getLinks().put(entry.getKey(), entry.getValue()));
```
mfbieber, line 57: Is this just a Map merge? If yes, why not use `item.getLinks().putAll(increment.getLinks());`

bonndan: It is putIfAbsent 😄

src/main/java/de/bonndan/nivio/input/Indexer.java
```diff
@@ -87,42 +88,45 @@ private void runResolvers(LandscapeDescription input, Landscape landscape) {
...
 
-        // apply template values to the items
-        new TemplateResolver().processTemplates(input, templatesAndTargets);
+        // apply template values to items
+        new TemplateResolver(logger).resolve(input);
 
-        // read special labels on items and assign the values to fields
-        new LabelToFieldProcessor(logger).process(input, landscape);
+        // resolve links on components to gather more data.
+        new LinksResolver(logger, linkHandlerFactory).process(input);
```
mfbieber, line 98: Shouldn't the method be named `resolve()`?

src/main/java/de/bonndan/nivio/input/Indexer.java
```diff
@@ -87,42 +88,45 @@ private void runResolvers(LandscapeDescription input, Landscape landscape) {
...
+        new LabelToFieldResolver(logger).resolve(input);
 
         // find items for relation endpoints (which can be queries, identifiers...)
-        new RelationEndpointResolver(logger).processRelations(input);
+        // KEEP here (must run late after other resolvers)
+        new RelationEndpointResolver(logger).resolve(input);
 
         // add any missing groups
-        new GroupResolver(logger).process(input, landscape);
+        new GroupProcessor(logger).process(input, landscape);
```
mfbieber, line 114: Since this is all about resolvers and processors. What is the difference between both of them? They both operate on the provided `LandscapeDescription input`. Is the result a modified input which is later applied to the `Landscape`?

src/main/java/de/bonndan/nivio/input/Indexer.java
bonndan: excellent observation, Sherlock!

src/main/java/de/bonndan/nivio/input/Indexer.java
bonndan: Maybe the naming can be discussed. Resolvers just modify and extended the input, but not not affect the final output.

src/main/java/de/bonndan/nivio/input/Indexer.java
bonndan: done.

src/main/java/de/bonndan/nivio/input/ItemDescriptionValues.java
```diff
@@ -13,32 +16,46 @@
...
+                    .filter(s -> !StringUtils.isEmpty(s) && !item.getProvidedBy().contains(s))
+                    .forEach(s -> item.getProvidedBy().add(s));
+        }
+
+        template.getRelations().forEach(item::addRelation);
+
+        template.getInterfaces().forEach(interfaceItem -> {
+            if (!item.getInterfaces().contains(interfaceItem))
+                item.getInterfaces().add(interfaceItem);
+        });
```
mfbieber, line 59: The contains check is not necessary. Can't this also be simpler like this: `item.getInterfaces().addAll(template.getInterfaces());`?

docs/source/input.rst
bonndan: ?

src/main/java/de/bonndan/nivio/input/LinksResolver.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+
+/**
+ * Resolves all links of all landscape components.
+ */
+public class LinksResolver {
```
mfbieber: Why does this not extend `Resolver`?

src/main/java/de/bonndan/nivio/input/ItemDescriptionValues.java
bonndan: done

src/main/java/de/bonndan/nivio/input/LinksResolver.java
bonndan: You're right, I have changed it based on your previous comment.

src/main/java/de/bonndan/nivio/input/Processor.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ *
+ */
+abstract class Processor {
+
+    protected final ProcessLog processLog;
+
+    protected Processor(ProcessLog processLog) {
+        this.processLog = processLog;
+    }
```
mfbieber, line 17: Do you use  an abstract class instead of an interface only because of the `ProcessLog` field?

src/main/java/de/bonndan/nivio/input/TemplateResolver.java
```diff
@@ -1,30 +1,44 @@
...
-    public void processTemplates(LandscapeDescription landscape, Map<ItemDescription, List<String>> templatesAndTargets) {
-        templatesAndTargets.forEach((landscapeItem, identifiers) -> applyTemplateValues(landscapeItem, identifiers, landscape));
+    @Override
+    public void resolve(LandscapeDescription landscape) {
+
+        Map<ItemDescription, List<String>> templatesAndTargets = new HashMap<>();
+        landscape.getSourceReferences().forEach(ref -> {
+            ref.getAssignTemplates().forEach((key, identifiers) -> {
+                ItemDescription template = landscape.getTemplates().get(key);
+                template.setName(null);
```
mfbieber, line 36: Why is the name set to `null`?

src/main/java/de/bonndan/nivio/input/dto/ComponentDescription.java
```diff
@@ -0,0 +1,26 @@
...
+
+    void setName(String name);
+
+    String getDescription();
+
+    void setDescription(String description);
+
+    String getOwner();
+
+    void setOwner(String owner);
```
mfbieber, line 25: Is there any `ComponentDescription` that will not have a name, description and owner when the object is constructed? We are working towards more inmutability in objects, I would not invite to use the setters through the interface and rather deprecate as much setters as possible ;)

src/main/java/de/bonndan/nivio/input/external/github/GitHubRepoHandler.java
```diff
@@ -0,0 +1,75 @@
...
+    public static final String DESCRIPTION = "github.description";
+
+    private final GitHub gitHub;
+
+    public GitHubRepoHandler(GitHub gitHub) {
+        this.gitHub = gitHub;
+    }
+
+    /**
+     * Loads the open issues from the repo.
```
mfbieber, line 37: This is cool 😻 

Shouldn't the description though be something like "Loads the open issues, open pull requests, the readme and the repositories description"?

src/main/java/de/bonndan/nivio/input/TemplateResolver.java
bonndan: Because names are more or less unique and I think it makes no sense to apply the same name to different items.

src/main/java/de/bonndan/nivio/input/external/springboot/CompositeHealthDeserializer.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.boot.jackson.JsonComponent;
+
+import java.io.IOException;
+import java.util.Map;
+
+@JsonComponent
+public class CompositeHealthDeserializer extends JsonDeserializer<JsonCompositeHealth> {
+
+    @Override
+    public JsonCompositeHealth deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
```
mfbieber, line 17: There is no test for this, is there? Also I didn't find one for the `JsonCompositeHealth`. Why not addind directly a test for this new functionality? The test should be rather simple now.

I started adding jacoco with https://github.com/dedica-team/nivio/pull/329 and from what I saw, we can really improve on test code coverage!

src/main/java/de/bonndan/nivio/input/Processor.java
bonndan: Yes, so far. We can change it to an interface and add the processlog as further argument to process/resolve.

docs/source/input.rst
mfbieber: "Nivio can load external data that **can not** be used directly to build landscapes, but is still valuable." -> "Nivio can load external data that **cannot** be used directly to build landscapes, but is still valuable. "

src/main/java/de/bonndan/nivio/input/dto/ComponentDescription.java
bonndan: Yes, it is very likely that these values become available later during processing only. I would leave the input DTOs (all what extends from ComponentDescription) as mutable as possible.

docs/source/input.rst
bonndan: done

src/main/java/de/bonndan/nivio/input/ComponentDescriptionValues.java
mfbieber: ok

src/main/java/de/bonndan/nivio/input/external/github/GitHubRepoHandler.java
bonndan: done

src/main/java/de/bonndan/nivio/input/Indexer.java
mfbieber: Why is an input modified if it does not have an effect on the final output?

I am just trying to understand the underlying mechanism ;)

src/main/java/de/bonndan/nivio/input/Processor.java
mfbieber: It's ok, I dont mind the abstract class. I think the only thing that "bothers" me is that I don't fully understand what the resolvers and the processors do and why - see comment above :)

src/main/java/de/bonndan/nivio/input/TemplateResolver.java
mfbieber: Isn't the name then `null` by default for a template? Why set it explicitly?

src/main/java/de/bonndan/nivio/input/Indexer.java
bonndan: First, the data is aggregated in the ComponentDescription objects. This is done sequentially by the different Resolvers. Then the processors do their work and start mapping the input to the final Landscape.

src/main/java/de/bonndan/nivio/input/TemplateResolver.java
mfbieber: I mean, annotate it as `@Nullable`, make the getter either return `@Nullable` or  @Nonnull` and an `Optional<String>` and the caller of the getter knows what to expect.

Also, the name is annotated with `@NotEmpty` in `ItemDescription`. I think `null` is not a valid value - is it?

mfbieber:
I think it is really cool to support Spring, GitHub and Sonarqube. Do we have examples for that, so that upon deployment of our demo nivio users can see them?

src/main/java/de/bonndan/nivio/input/Processor.java
bonndan: Until a few days ago I was not aware of that, too. I just observed that some of the classes in Indexer only worked on the input objects, while others only/mostly worked on the output.

src/main/java/de/bonndan/nivio/input/TemplateResolver.java
bonndan: ItemDescription is misused as template. And perhaps you are right: there could exist items without a name (not without identifier, however).

bonndan: I have updated the interface and added `@nullable` annotations

src/main/java/de/bonndan/nivio/input/external/springboot/CompositeHealthDeserializer.java
bonndan: It's tested indirectly in `SpringBootHealthHandlerTest`. Although you're right, I am not very motivated to invest my time in an JsonParser/TreeNode/bla etc. deep dive.

bonndan:
> I think it is really cool to support Spring, GitHub and Sonarqube. Do we have examples for that, so that upon deployment of our demo nivio users can see them?

Sonarqube ist not implemented yet. I'll add examples for the other sources.

src/main/java/de/bonndan/nivio/input/Indexer.java
mfbieber: Okay thanks, now I understand!

src/main/java/de/bonndan/nivio/input/dto/ComponentDescription.java
mfbieber: Okay

Issue #302, opened 2020-09-22
Improve search UX


bonndan:
Try out the search and make proposals how to improve it.

Hints:

* search syntax is Lucene style, which is precise but uncommon if compared to Google search
* where should the results appear?
* Should the results disappear after click?

bonndan:
From https://github.com/dedica-team/nivio/pull/342

- [ ] find results regardless of upper and lowercase input
- [ ] find also partial words, e.g. with "word" I would like to already get "wordpress". Maybe one could match once an user has typed 3 chars
- [ ] "emailer" does not find the emailer. This is most probably a bug?

bonndan:
#376 improves search behavior

PR #342, opened 2020-12-21
Using materialui

bonndan:
Front page
image

Dashboard
image

Map
image


src/main/app/src/Components/Landscape/Map/Map.tsx
```diff
@@ -40,73 +38,73 @@ interface SVGData {
  * @param identifier Landscape Identifier
  * @param onItemClick Handler for our label click
  */
-const Map: React.FC<Props> = () => {
+const Map: React.FC<Props> = ({ setSidebarContent, setFindFunction, setPageTitle }) => {
```
mfbieber, line 41: This doc is outdated (onItemClick does not appear in function signature) and "choosen" is misspelled

bonndan:
Front page made of standard cards
image

Map view with item details
image

- dashboard has been removed, is now part of the map view (at the bottom)
- item details contain more info, especially relation data (clickable)

mfbieber:
The cards look great and clean! Good job!

Also the map view is nice. Just two suggestions, I think the first one is more essential than the second:
nivio
1.  The colors of the buttons at the bottom do not match wich the group colors. I know that it is the same color, but the group backgrounds are transparent. I would suggest to add the same transparancy to the buttons.
2.  I would probably add a bottom bar similiar to the top one for those buttons, so that they are not overlaying the map. In order for the colors to work fine, probably a dark color (and not the green color of the AppBar) is beneficial

mfbieber:
One more thing about the icons of the items:
nivio2
On the right side, the picture for the wordpress-web (Apache) item is just the same as the left one and scaled down only a slight bit.

This looks odd. I know that it is not readable if scaled down to fit into the circle, but I think it is better than cutting it like it is now.

Was this on purpose or did this happen by accident?

src/main/app/src/Components/Landscape/Map/Map.tsx
```diff
@@ -270,6 +267,16 @@ const Map: React.FC<Props> = () => {
...
         />
+        <Toolbar className={classes.bottomBar}>
+          {landscape && assessments && (
+            <Dashboard
+              findItem={findItem}
+              setSidebarContent={setSidebarContent}
+              landscape={landscape}
+              assessments={assessments}
+            />
+          )}
```
mfbieber, line 278: I thought the dashboard had been removed. What is this then? Is this the container for the buttons on the bottom of a map?

mfbieber: I would rename this to something like e.g. StatusLegend or something that better represents what this actually is (also `DashboardLayout`. In addition, 

I just saw that this is just the Toolbar I was talking about in one of the other comments (I wouldn't allow the map to be visible below it).

src/main/app/src/App.tsx
```diff
@@ -1,33 +1,61 @@
...
+  const [findFunction, setFindFunction] = useState<Function>();
+
+  const ff = useCallback(
+    (args) => {
+      if (findFunction) findFunction(args);
+      else console.warn('find function not set');
+    },
+    [findFunction]
+  );
+
```
mfbieber, line 22: Is `findFunction` a function for the search field? I would name it `searchFunction`

src/main/app/src/Components/Landscape/Map/Map.tsx
```diff
@@ -177,6 +167,23 @@ const Map: React.FC<Props> = () => {
...
+        if (findItem) {
+          setFindFunction(() => findItem);
+        }
+      });
+
+      get(`/assessment/${identifier}`).then((response) => {
+        setAssessments(response);
+      });
+    }
+  }, [identifier, setPageTitle, setFindFunction, findItem, landscape]);
```
mfbieber, line 185: Do we display any information to the user if a call to the backend is not succesful? We could add something to the `APIClient` like
````
axios.get('endpoint')
    .then((response) => {})
    .catch((error) => {
        console.log(error); //better something with information for the user
    })
````

File: src/main/app/src/Components/Landscape/Map/MapRelation/MapRelation.tsx [hunk 5]
```diff
@@ -10,49 +13,51 @@ interface Props {
 
 /**
  * Returns a choosen Map Relation
```
mfbieber, line 15: It is "chosen", not "choosen"

src/main/app/src/Components/Landscape/Map/MapRelation/MapRelation.tsx
```diff
@@ -10,49 +13,51 @@ interface Props {
...
- * @param element Choosen SVG Element from our Landscape Component
+ * TODO work with items and relation objects
```
mfbieber, line 16: What has to be done with this `TODO`?

And I noticed that it is really difficult to click on the relation. I am quite sure, that an user would not have a clue that the relations are clickable. I just came across the fact that it is clickable, because I looked up how `MapRelation` is used in the code.

1. The lines must be thicker in order to be clickable properly
2. A hint or something must exist to invite the user to click them, e.g. also change the mouse to the hand symbol when hovering over the (thicker) lines

src/main/app/src/Components/Landscape/Map/MapRelation/MapRelation.tsx
```diff
@@ -10,49 +13,51 @@ interface Props {
...
-      </div>
-    </div>
+    <Card className={classes.card}>
+      <CardHeader title={title} className={classes.cardHeader}/>
+      <CardContent>
+        <span>Type: {type}</span>
+        <br />
+        <br />
+        Source:
+        <br />
```
mfbieber, line 37: Instead of using this elements to style the text, consider using `Typography` from Material UI. I also saw it in the `Events.tsx`.

src/main/app/src/Components/Landscape/Modals/Group/Group.tsx
```diff
@@ -1,87 +1,104 @@
...
-  fullyQualifiedGroupIdentifier: string;
+  group: IGroup;
+  assessments: IAssessment;
   findItem?: (fullyQualifiedGroupIdentifier: string) => void;
   findGroup?: (fullyQualifiedGroupIdentifier: string) => void;
-  onAssessmentClick?: (fullyQualifiedItemIdentifier: string) => void;
 }
 
 /**
  * Returns a choosen Landscape group if informations are available
```
mfbieber, line 21: Here it is also "chosen" and not "choosen"

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
```diff
@@ -1,116 +1,204 @@
...
+  const classes = componentStyles();
+  let relations: ReactElement[] = [];
 
   useEffect(() => {
-    get(`/api/${fullyQualifiedItemIdentifier}`).then((item) => {
+    const loadAssessment = (item: IItem) => {
+      const landscapeIdentifier = item ? item.fullyQualifiedIdentifier.split('/') : [];
+
+      if (landscapeIdentifier[0]) {
+        //TODO load direct to with fqi in path
```
mfbieber, line 37: I also don't understand this `TODO`. I just can't parse the sentence :) What needs to be done here?

src/main/app/src/Components/Landscape/Modals/Group/Group.tsx
```diff
@@ -1,87 +1,104 @@
...
  */
-const Group: React.FC<Props> = ({
-  fullyQualifiedGroupIdentifier,
-  findItem,
-  findGroup,
-  onAssessmentClick,
-}) => {
-  const [group, setGroup] = useState<IGroup | undefined>();
-  const [assessment, setAssessment] = useState<IAssessmentProps[] | undefined>(undefined);
+const Group: React.FC<Props> = ({ group, assessments, findItem, findGroup }) => {
```
mfbieber, line 23: I noticed that in the map, it is not possible to click on the group background to focus on that group. I would have expected to get a focus on it and to have the group information displayed at the "sidebar"

src/main/app/src/Components/Landscape/Map/Map.tsx
```diff
@@ -36,79 +51,59 @@ interface SVGData {
...
- * @param identifier Landscape Identifier
- * @param onItemClick Handler for our label click
+ * Displays a chosen landscape as interactive SVG
+ *
+ * @param setSidebarContent function to set sidebar/drawer content
+ * @param setFindFunction function to use to find an item. make sure to pass an anon func returning the actually used function
+ * @param setPageTitle can be used to set the page title in parent state
  */
-const Map: React.FC<Props> = () => {
+const Map: React.FC<Props> = ({ setSidebarContent, setFindFunction, setPageTitle }) => {
```
mfbieber, line 60: The sidebar is more like a card. I would call it `InformationCard`or something else than "sidebar". I think sidebar is actually a bar at the right or left side of the page.

src/main/app/src/Components/Landscape/Search/Search.tsx
```diff
@@ -31,39 +43,50 @@ interface ILabelValue {
...
+    (searchTerm: string, identifier: string) => {
+      if (searchTerm.length < 2) return;
+
+      get(
+        '/api/landscape/' +
+          identifier +
+          '/search/' +
+          encodeURIComponent(searchTerm)
+            .replace(/[!'()]/g, escape)
+            .replace(/\*/g, '%2A')
```
mfbieber, line 73: I like the search field 👍 

Unfortunately, it does not work well all the time. This should be better (backend functionality):

- find results regardless of upper and lowercase input
- find also partial words, e.g. with "word" I would like to already get "wordpress". Maybe one could match once an user has typed 3 chars
- "emailer" does not find the emailer. This is most probably a bug?

src/main/app/src/Components/Navigation/Navigation.tsx
```diff
@@ -1,61 +1,70 @@
...
+    },
+    menuIcon: {
       color: 'rgba(255, 255, 255, 0.75)',
     },
   })
 );
 
+interface Props {
+  appBarClass: string;
+  setSidebarContent: Function;
```
mfbieber, line 32: See my comment above about the naming of the "sidebar", that in my opinion is a card showing information and not a sidebar :)

src/main/app/src/Components/Landscape/Map/MapRelation/MapRelation.tsx
bonndan: The TODO says that there are objects for items and relations available (loaded via API), but these are not used in here.

Regarding map relations: https://github.com/dedica-team/nivio/issues/348

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
bonndan: I meant it is not necessary to split strings since each item has a fullyQualifiedIdentifier that can be used directly as API call path element. Also, the complete data has already been loaded (but not by the search component)

src/main/app/src/Components/Landscape/Modals/Group/Group.tsx
bonndan: https://github.com/dedica-team/nivio/issues/349

src/main/app/src/Components/Navigation/Navigation.tsx
bonndan: My idea was that the sidebar could hold any information, although now mostly single cards appear. However, if you click the "log" button on the front page or the "more/extend" symbol next to the search field, other content that cards appears.

src/main/app/src/Components/Landscape/Map/Map.tsx
bonndan: fixed

src/main/app/src/Components/Landscape/Search/Search.tsx
bonndan: I've quoted your remarks in https://github.com/dedica-team/nivio/issues/302

bonndan:
> One more thing about the icons of the items:
> nivio2
> On the right side, the picture for the wordpress-web (Apache) item is just the same as the left one and scaled down only a slight bit.
> 
> This looks odd. I know that it is not readable if scaled down to fit into the circle, but I think it is better than cutting it like it is now.
> 
> Was this on purpose or did this happen by accident?

We have to style/tweak the "Avatar" Material UI component.

src/main/app/src/Components/Landscape/Map/Map.tsx
bonndan: As mentioned the sidebar indeed is a container for content, mostly cards, and goes from top to bottom 😄

src/main/app/src/Components/Landscape/Modals/Group/Group.tsx
bonndan: fixed

src/main/app/src/Components/Landscape/Map/MapRelation/MapRelation.tsx
bonndan: fixed

src/main/app/src/App.tsx
bonndan: It actually is "locateFunction" to scroll to the given item in the map. I will rename it.

bonndan:
image

This is better... (taken from a material UI example)

src/main/app/src/Components/Landscape/Map/Map.tsx
bonndan: See https://github.com/dedica-team/nivio/issues/350

bonndan:
also closes #349

src/main/app/src/Components/Navigation/Navigation.tsx
mfbieber: Okay, so the sidebar contains cards. Alright ;)

src/main/app/src/Components/Landscape/Map/MapRelation/MapRelation.tsx
mfbieber: The hand symbol is working, nice!

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
mfbieber: So we could just do `get(`/assessment/${landscapeIdentifier}`)` directly and our backend already supports that?

Or what has to be done to be able to do that call?

Regarding the already loaded data: do you mean that this call is not really necesary because another call to the backend already retrieved the data this call would retrieve?

mfbieber:
I went through your recent changes: 👍

Issue #164, opened 2020-06-17
Start out of the box without configuration file


bonndan:
Instead of requiring a configuration file, nivio could also start with a single landscape configured using environment variables (e.g. kubernetes api).

Example: https://docs.traefik.io/reference/static-configuration/env/

- [ ] find out how
- [ ] determine good variables
- [ ] document

Matthimatiker:
>  the sequence of variables should be kept (so that overwriting rules are clear)

I guess this could be handled by Spring Boot. Please check if that system is sufficient, before implementing another config layer.

-------

On a higher abstraction level, the Twelve Factor App config principle might be interesting for this task: https://12factor.net/config

PR #351, opened 2021-01-04
Gitlab link resolver

### LINKED ISSUE #347: Implement GitLab repo link handler
Similar to GitHub handler, scan a GitLab repo.

bonndan:
closes #347 

also addresses #164 , configurable env vars in class and generated docs


src/main/java/de/bonndan/nivio/input/Seed.java
```diff
@@ -13,30 +13,27 @@
...
-    private String seed;
 
-    public Seed() {
-    }
+    private final String seed;
 
-    public Seed(String seed) {
-        this.seed = seed;
+    public Seed(Optional<String> seed) {
+        this.seed = seed.orElse(null);
```
mfbieber, line 28: This `Optional` is the return value of the `ConfigurableEnvVars` enums method `value()`:
````
public Optional<String> value() {
   return Optional.ofNullable(System.getenv(name()));
}
````
If this is transformed to `null` anyways, why bother using the `Optional`? I think this only makes sense, if the `Optional`is really used to handle some kind of state.

File: src/main/java/de/bonndan/nivio/config/ConfigurableEnvVars.java [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Returns the string value of the configuration variable if present in the environment.
+     *
+     * @return env value
+     */
+    public Optional<String> value() {
+        return Optional.ofNullable(System.getenv(name()));
+    }
```
mfbieber, line 57: I found the naming a little bit misleading at first, because I expected the String value of the enum constant. But it is indeed a value, so I don't know - just wanted to comment.

src/main/java/de/bonndan/nivio/config/ConfigurableEnvVars.java
```diff
@@ -0,0 +1,59 @@
...
+    GITHUB_PASSWORD("GitHub password (for username/pw login)"),
+    GITHUB_OAUTH("To connect to GitHub via Personal access token"),
+    GITHUB_JWT("To connect via JWT token as a GitHub App:"),
+
+    GITLAB_HOST_URL("The full URL to the GitLab API, e.g. http://your.gitlab.server.com"),
+    GITLAB_PERSONAL_ACCESS_TOKEN("the personal token to access the GitLab API (optional)"),
+    GITLAB_USERNAME("GitLab OAuth login username (optional, if used password is also required)"),
+    GITLAB_PASSWORD("GitLab OAuth login password (optional)"),
+
+    KUBERNETES_MASTER("K8s master url (optional), all variables from https://github.com/fabric8io/kubernetes-client#configuring-the-client can be used"),
```
mfbieber, line 26: So nivio can work with all the variables from the provided link for k8s out of the box? Did I understand that correctly?

src/main/java/de/bonndan/nivio/input/external/RepositoryLinkHandler.java
```diff
@@ -0,0 +1,26 @@
...
+    String ICON = "repo.icon";
+
+    @NonNull
+    default String getRepoName(Link link) {
+        try {
+            String s = link.getHref().toString().split(link.getHref().getHost())[1];
+            s = StringUtils.trimTrailingCharacter(s, '/');
+            return StringUtils.trimLeadingCharacter(s, '/');
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to parse github url " + link.getHref() + ": " + e.getMessage());
```
mfbieber, line 22: This default method is used both for GitHub and GitLab. I think the message should not be specific to GitHub ;)

src/main/java/de/bonndan/nivio/input/Seed.java
mfbieber: Okay, I saw that in e.g. GitLabHandler, you are really using isPresent checks.

mfbieber (commented):
**Nice that we can connect to GitLab and GitHub! 🎉** 

- Do we have example repositories to show in our demo? 
- I guess you tested it with real remote repositories?


**Regarding the environment variables:** 

If I am really to use nivio for my GitLab repositories, k8s cluster and sonar together, I would instantly be annoyed by having to set all those environment variables. 

Did you consider working with application.properties files and use the system environment variables as a fallback?


**Kubernetes support:, is this really resolved (https://github.com/dedica-team/nivio/issues/164)?**

>  find out if nivio is running in an k8s cluster by examining variables, then start the k8s scanner

No, #164 would remain open. 

**Sequence of variables and Matthias' comment (https://github.com/dedica-team/nivio/issues/164)**

> the sequence of variables should be kept (so that overwriting rules are clear)

When would this be relevant with the current implementation as system environment variables?

src/main/java/de/bonndan/nivio/config/ConfigurableEnvVars.java
bonndan: Yes. Last time I tried it worked :)

src/main/java/de/bonndan/nivio/config/ConfigurableEnvVars.java
bonndan: I thought about it when I wrote it, but had no better idea.

src/main/java/de/bonndan/nivio/input/external/RepositoryLinkHandler.java
bonndan: fixed

bonndan:
> **Nice that we can connect to GitLab and GitHub! 🎉**
> 
> * Do we have example repositories to show in our demo?
> * I guess you tested it with real remote repositories?

I have a hidden repo at GitLab.

> **Regarding the environment variables:**
> 
> If I am really to use nivio for my GitLab repositories, k8s cluster and sonar together, I would instantly be annoyed by having to set all those environment variables.
> 
> Did you consider working with application.properties files and use the system environment variables as a fallback?

This would stand against 12-factor stuff and immutable services (in terms of using nivio with docker). Or do you mean this scary stuff: https://docs.spring.io/spring-boot/docs/2.1.9.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-application-property-files ?

However, you'd have to collect the values anyway once, be it env vars or a config file.

> 
> **Kubernetes support:, is this really resolved (#164)?**
> 
> > find out if nivio is running in an k8s cluster by examining variables, then start the k8s scanner
> 
> **Sequence of variables and Matthias' comment (#164)**
> 
> > the sequence of variables should be kept (so that overwriting rules are clear)
> 
> When would this be relevant with the current implementation as system environment variables?

Actually I am not sure anymore what "sequence of variables" was supposed to mean 🙈

mfbieber:
> This would stand against 12-factor stuff and immutable services (in terms of using nivio with docker). Or do you mean this scary stuff: https://docs.spring.io/spring-boot/docs/2.1.9.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-application-property-files ?

Yes, this is what I meant. This is not so horrible. I've seen it in praxis several times and it really depends on how that explodes if it gets scary. I would have allowed the possibility to use it.

> However, you'd have to collect the values anyway once, be it env vars or a config file.

Sure. If I have 100 values, I prefer the file ;)

> I have a hidden repo at GitLab.

And what about showing the support of this feature in our demo deployed nivio?

bonndan:
see #358

mfbieber:
> see #358 

👍 

> The "inout" landscape item "nivio" has a link to its GitHub repos and reads the values (readme content for instance). Is that sufficient?

I can't see it in https://nivio-demo.herokuapp.com/#/landscape/inout right now, but I guess after the branch has been merged?

Then, go ahead ;)

PR #367, opened 2021-01-12
Fixing pointer mouse cursor for triangles in relations

mfbieber:
- made dashes "butt" instead of "round"
- made lines and triangles thicker 
- added cursor pointer to triangles by setting the attribute to the .map g.relation (the triangles do not contain a path tag)

I think now it looks a little bit better:
nivio5


bonndan (approved):
Works on my machine :)

Issue #365, opened 2021-01-12
Focus / Jump to is broken


bonndan:
In the "inout" map view nearly no click on "focus" scrolls properly to center the item.

PR #369, opened 2021-01-14
[#365] first attempt to improve centering

bonndan:


mfbieber (approved):
Centering looks better now, altough not perfect centered - or?

bonndan:
Yes, there is still room for improvement. Perhaps we need to have a closer look at viewer dimensions vs svg dimensions.

PR #376, opened 2021-01-17
375 fix search locate

### LINKED ISSUE #375: jump-to/locate function broken after search
Clicking the focus icon in the search results does not scroll to an item's location. Maybe related to landscape change (old function still active).

bonndan:
closes #375

### DIFF
--- a/src/main/app/src/Components/Landscape/Map/Map.tsx
+++ b/src/main/app/src/Components/Landscape/Map/Map.tsx
@@ -188,7 +188,7 @@ const Map: React.FC<Props> = ({ setSidebarContent, setLocateFunction, setPageTit
     if (locateComponent) {
       setLocateFunction(() => locateComponent);
     }
-  }, [value, setLocateFunction, locateComponent]);
+  }, [setLocateFunction, locateComponent]);
 
   useEffect(() => {
     let timeout: NodeJS.Timeout;
--- a/src/main/app/src/Components/Landscape/Map/Map.tsx
+++ b/src/main/app/src/Components/Landscape/Map/Map.tsx
@@ -23,7 +23,7 @@ import StatusBar from '../Dashboard/StatusBar';
 import { IAssessment, ILandscape } from '../../../interfaces';
 import { getGroup, getItem } from '../Utils/utils';
 import Group from '../Modals/Group/Group';
-import MapUtils from "./MapUtils";
+import MapUtils from './MapUtils';
 
 interface Props {
   setSidebarContent: Function;
@@ -60,7 +60,6 @@ const Map: React.FC<Props> = ({ setSidebarContent, setLocateFunction, setPageTit
 
   const [isFirstRender, setIsFirstRender] = useState(true);
 
-
   const locateComponent = useCallback(
     (fullyQualifiedItemIdentifier: string) => {
       const element = document.getElementById(fullyQualifiedItemIdentifier);
@@ -75,7 +74,7 @@ const Map: React.FC<Props> = ({ setSidebarContent, setLocateFunction, setPageTit
         }
       }
     },
-    [value]
+    [setValue, value]
   );
 
   const onItemClick = (e: MouseEvent<HTMLElement>) => {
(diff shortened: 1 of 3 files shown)


src/main/app/src/Components/Landscape/Map/Map.tsx
```diff
@@ -75,7 +74,7 @@ const Map: React.FC<Props> = ({ setSidebarContent, setLocateFunction, setPageTit
         }
       }
     },
-    [value]
+    [setValue, value]
```
mfbieber, line 77: As I understood the deps of the `useCallback()` hook, these are values that change upon which the hook is triggered (https://reactjs.org/docs/hooks-reference.html#usecallback). `setValue` never changes (does it?). 

I also tested removing it, it makes no difference for the focus function.

src/main/app/src/Components/Landscape/Map/Map.tsx
```diff
@@ -177,16 +176,19 @@ const Map: React.FC<Props> = ({ setSidebarContent, setLocateFunction, setPageTit
...
       });
     }
-  }, [identifier, setPageTitle, setLocateFunction, locateComponent, landscape]);
+  }, [identifier, setPageTitle, landscape]);
+
+  useEffect(() => {
+    if (locateComponent) {
+      setLocateFunction(() => locateComponent);
+    }
+  }, [value, setLocateFunction, locateComponent]);
```
mfbieber, line 191: Just as a question, are these all deps that are needed?

src/main/app/src/Components/Landscape/Search/Search.tsx
```diff
@@ -63,7 +63,7 @@ const Search: React.FC<PropsInterface> = ({locateFunction, setSidebarContent, ..
   const search = useCallback(
     (searchTerm: string, identifier: string) => {
       if (searchTerm.length < 2) return;
-
+      setHasChange(false);
```
mfbieber, line 66: Wheter it is here or after the `get()` should not really matter - does it?

src/main/app/src/Components/Landscape/Map/Map.tsx
mfbieber: At least `value` is not needed in line 191.

src/main/app/src/Components/Landscape/Map/Map.tsx
mfbieber: Update: the deps should be values that are used within the hook. `setValue` is not used in this hook, we can remove it.

mfbieber (commented):
Focusing works ;)

src/main/app/src/Components/Landscape/Search/Search.tsx
bonndan: Yes, here it prevents further requests because the ui is rendered several times before the response is available.

src/main/app/src/Components/Landscape/Search/Search.tsx
```diff
@@ -73,20 +73,22 @@ const Search: React.FC<PropsInterface> = ({locateFunction, setSidebarContent, ..
...
-        setSidebarContent(searchResult);
-        setHasChange(false);
-      });
-    },
-    [results, setSidebarContent, locateFunction]
-  );
+        />
+    ));
+    setSidebarContent(searchResult);
+  }, [results, setSidebarContent, locateFunction]);
```
mfbieber, line 91: `setResults(result)` triggers a state change in the `Search` component and provides the new `results` to the component after a new render.

` const searchResult = results.map()` operates on the `results` available in the component. Since the new `results` are only available on the new render, we lagged behind one input character.

Now, located on an own `useEffect()`, the `searchResult` is passed on to the `setSidebarContent` function correctly, which triggers again a new render.

mfbieber (commented):
I moved the `setSidebarContent()` call to an own `useEffect()`. This repairs the unecessary typing of one more character in the search field.

src/main/app/src/Components/Landscape/Map/Map.tsx
bonndan: fixed

src/main/app/src/Components/Landscape/Search/Search.tsx
bonndan: excellent 👍

src/main/app/src/Components/Navigation/Navigation.test.tsx
```diff
@@ -6,7 +6,7 @@ import { MemoryRouter } from 'react-router-dom';
 it('should render Navigation component', () => {
   const { getByText } = render(
     <MemoryRouter>
-      <Navigation />
+      <Navigation setSidebarContent={() => {}} appBarClass={'foo'} locateFunction={() => {}} />
```
mfbieber, line 9: 😆 I was just fixing it too!

mfbieber (commented):
👍

PR #390, opened 2021-01-24
355 improve relation resolver

### LINKED ISSUE #355: Improve magic relation resolver
Splitting and parsing label keys leads to too many false positives.

Produce hints.

- [ ] detect and parse DSNs to produce hints: e.g. jdbc, hostname, type (database)
- [ ] can global aliases (alternative identifiers)  be used?

bonndan:
* removes false-positive relations found by "magic" resolver (that was the purpose of the issue, closes #355)
* items have "address" field (technical address) which is indexed
* all labels are examined (except for configured blacklist, as before)
* refactoring of search/indexing: removes duplicate index implementation, extracted lucene search index to own class


mfbieber:
When navigating from the initial screen to the landscape example, the processing event log is still displayed. If you zoom in, it does not look very good on the map:

nivio_process_log

I guess that the wanted behavior would be that the process log is not displayed, correct?

Is the display log located in the sidebar? I think its contents must be cleared upon navigation.

bonndan:
> Is the display log located in the sidebar? I think its contents must be cleared upon navigation.

done.

PR #392, opened 2021-01-25
Dedica landscape

### LINKED ISSUE #393: Include all svg map images
If not all images are embedded as data urls

* the svg is not portable
* the preview breaks on the front page

- [x] embed logo
- [x] embed "fill" urls
- [x] embed external icon urls

bonndan:
Also closes #393 

Dedica landscape is contained in DEMO mode. Contains fixes for "fill" feature.

image


mfbieber (approved):
Looks good. 

The only thing I noticed is that maybe we should make the group background a little bit larger, so that the names of the items are contained in it (this is not specific to this PR, I noticed it in the other landscape examples too).

Issue #436, opened 2021-02-15
Highlight changes after input data update


bonndan:
- [x] log which items have been updated, added or deleted
- [ ] Display changes in newly loaded map

PR #443, opened 2021-02-16
435 feedback cycle

### LINKED ISSUE #435: Improve landscape update feedback loop
As a user trying out nivio I need to see input data changes/updates immediately. Otherwise I have the impression that the system is broken or I did something wrong.

- [x] ensure the file based change observers pick up changes immediately
- [x] ensure websocket based change notification is still working

- see #322

bonndan:
Fixes websockets, improves notification behaviour
closes #435


src/main/app/src/Components/Notification/Notification.tsx
```diff
@@ -19,34 +19,46 @@ const Notification: React.FC = () => {
...
+  }, [socketUrl, setClient, setConnected]);
+
+  useEffect(() => {
+    if (!client || subscription)
+      return;
+    if (!connected)
+      return;
+    const stompSubscription = client.subscribe('/topic/events', (message) => {
+      const notificationMessage: INotificationMessage = JSON.parse(message.body);
+      console.log(notificationMessage);
```
mfbieber, line 46: Is this `console.log()` on purpose?

src/main/app/src/Components/Notification/Notification.tsx
```diff
@@ -19,34 +19,46 @@ const Notification: React.FC = () => {
...
+      //debug: (msg) => console.log(msg),
+      onConnect: (receipt) => {
+        setConnected(true);
       },
-    })
-  );
+    });
+    client1.activate();
+    setClient(client1);
+  }, [socketUrl, setClient, setConnected]);
```
mfbieber, line 37: You should't need the set functions as deps, also down below in line 61

src/main/app/src/Components/Notification/Notification.tsx
```diff
@@ -19,34 +19,46 @@ const Notification: React.FC = () => {
...
-        });
+  const [client, setClient] = useState<Client>();
 
-        subscriptions.push(eventSubscription);
-        setSubscriptions(subscriptions);
+  useEffect(() => {
+    const client1 = new Client({
+      brokerURL: socketUrl,
+      //debug: (msg) => console.log(msg),
+      onConnect: (receipt) => {
```
mfbieber, line 31: the receipt parameter is not used

src/main/app/src/Components/Notification/Notification.tsx
```diff
@@ -19,34 +19,46 @@ const Notification: React.FC = () => {
...
+          key: new Date().getTime(),
+          landscape: notificationMessage.landscape,
+          level: notificationMessage.level,
+        };
+        setSnackPack((prevArray) => [...prevArray, snackPackMessage]);
+        setOpen(true);
+      }
+    });
+    setSubscription(stompSubscription);
+  }, [client, connected, subscription, setSubscription])
```
mfbieber, line 61: Why is a change in either `client`, `connected` or `subsctiption` needed to trigger this effect? I mean, besides obviously that it does, since they are in the deps array. How do these even change? I would have expected them to be created once (as it was before) during the first render.

mfbieber: Is it because of achaning socketUrl, that the client needs to be recreated?

Why does this URL change?

src/main/java/de/bonndan/nivio/input/SourceReferencesResolver.java
```diff
@@ -2,37 +2,75 @@
...
-                log.warn("Failed to resolve source reference '" + ref.getUrl() + "' (" + ref.getFormat() + ")", ex);
+                String message = ex.getMessage();
+                if (ex instanceof ReadingException) {
+                    message += ": " + getCauseMessage(ex.getCause());
+                }
+                log.error(message);
+                eventPublisher.publishEvent(new ProcessingErrorEvent(landscapeDescription, ex));
+                landscapeDescription.setIsPartial(true);
+            } catch (RuntimeException ex) {
+                log.warn(ex.getMessage());
```
mfbieber, line 60: This is basically just to have mor detailed error messages?

mfbieber:
I have seen that we get a nofitification, that the landscape has changed, but it takes a while until that message appears (does it check for changes twice a minute?). 

I am editing the dedica.yml in notepad and saving explicitly (not in IntelliJ, see https://github.com/dedica-team/nivio/issues/435#issuecomment-787501807). 

Also, the 'show map' button does nothing and I still have to reload the page to see the changes.

notification

From the changes in the code in this PR, I don't really understand how these should fix the underlying issue. 

The UI needs to re-render the specific components on a change. In which component are they rendered - we need to look into that component. The notification seems to work now (except for the button).

As for the changes in the backed, maybe I overlooked it, but most of the changes seemed to be to me for better error handling. Was the newer Spring Version responsible for the "fix"?

src/main/app/src/Components/Notification/Notification.tsx
bonndan: no :)

src/main/java/de/bonndan/nivio/input/SourceReferencesResolver.java
bonndan: Yes. We are forwarded JSON parse errors to the frontend.

bonndan:
Yes, the show map button is not a good solution. However, we are not able yet to notify the frontend WHICH component has been changed. This PR is about the feedback loop, what you're describing is https://github.com/dedica-team/nivio/issues/436

mfbieber:
> Yes, the show map button is not a good solution. However, we are not able yet to notify the frontend WHICH component has been changed. This PR is about the feedback loop, what you're describing is #436

The button does nothing for me :-/

src/main/app/src/Components/Notification/Notification.tsx
bonndan: reverted

bonndan:
We still need that other PR https://github.com/dedica-team/nivio/pull/445 to complete reloading on changes, but now we have file system observers again

mfbieber:
Ok, so the file watcher is working, but I still notice things. 

- For example, `SVGDocument` logs two to three times, if we edit the file in IntelliJ. Obviously, we should debounce the response, in case of many fast subsequent events. 
- The log message says "rendering", does it really render? Or does the frontend render?
- The notification in the frontend is fine, as long as the user does not have to click it away. 
- Also, it pops up several times. Maybe as often as we could debounce in the backend?
- The frontend is too broken in general. Zoom does not work (let's get rid of buttons that show broken functionality). 
- The message in the notification popup says "index landscape", does it refer to "Landscape was indexed"? And what should a user do with that information? And what is a landscape compared to a map?

Also, I am wondering if we need the file system watching mechanism at all. I think most users would run nivio in a distributed system, possibly a cluster anyways. Then, the config files will e.g. be either in a repository or e.g. in a nginx used to serve them. Nivio is capable to retrieve the files from those (which is great), they could just use a webhook mechanism in nivio (e.g. a POST endpoint) to notify nivio that there has been a change in those. By this, you could. e.g. modify your config file in your repository and upon running the CI/CD pipeline, simply run a curl POST to nivio to notify nivio about the change. Later on, we could implement the possibility to provide the config directly with the POST body.

I am also suggesting this also, because this file watching mechanism (as technically cool it may be and I know we have spent some effort on implementing it), has been unreliable and breaking several times now. I also don't like that we don't really understand why it doesn't work on macOS and that we just increase a timeout and so on. 

In general, let's remove everything that is not working reliably as we want it to, that adds complexity we don't necessarily need, and most importantly, that we don't fully understand. Let's focus on those things that we can control and do well.

mfbieber:
@bonndan: can you also briefly point out what exactly in this PR "fixes websockets", so that I can more easily understand?

mfbieber:
The main improvement I was expecting from this PR is that we don't have to reload (e.g. hit F5) in the browser to see a change. For me this is clearly not working.

As I just got feedback from a user: they would like to use nivio (post COVID) in a teamspace tv to show a dashboard. They can't click on reload or hit any buttons to remove notifications.

bonndan:
> @bonndan: can you also briefly point out what exactly in this PR "fixes websockets", so that I can more easily understand?

The browser was not able to connect to the server via websockets, since the server did not respond properly (for reasons I don't know). Websockets are one basic requirement for the automatic update. Upgrading Spring Boot to a newer version fixed it (lucky guess).

bonndan:
As I wrote above, the automatic map update is not yet implemented and not scope of #435 . We need the changes in this PR and the ones in #445 to get there.

Issue #437, opened 2021-02-15
Ensure that all source references are loaded and that groups can be added in source references


bonndan:
Source references should behave exactly like landscape config yaml files (except that further source references cannot be linked).

- [x] ensure that groups can be configured
- [x] notify user of broken configurations

PR #448, opened 2021-02-20
437 source reference groups

bonndan:
#437 groups and templates can be set in source references

### DIFF
--- a/src/main/java/de/bonndan/nivio/input/SourceReferencesResolver.java
+++ b/src/main/java/de/bonndan/nivio/input/SourceReferencesResolver.java
@@ -26,10 +26,9 @@ public class SourceReferencesResolver {
         URL baseUrl = URLHelper.getParentPath(landscapeDescription.getSource()).orElse(null);
         landscapeDescription.getSourceReferences().forEach(ref -> {
             try {
-                InputFormatHandler factory = formatFactory.getInputFormatHandler(ref);
-                landscapeDescription.addItems(factory.getDescriptions(ref, baseUrl));
+                formatFactory.getInputFormatHandler(ref).applyData(ref, baseUrl, landscapeDescription);
             } catch (ProcessingException ex) {
-                log.warn("Failed to resolve source reference '" + ref.getUrl() + "' (" + ref.getFormat() + ")", ex);
+                log.warn(String.format("Failed to resolve source reference '%s' (%s)", ref.getUrl(), ref.getFormat()), ex);
                 landscapeDescription.setIsPartial(true);
             }
         });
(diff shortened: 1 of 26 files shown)


src/test/java/de/bonndan/nivio/input/dto/InputFormatHandlerNivioTest.java
```diff
@@ -87,16 +93,35 @@ public void readServiceAndInfra() {
...
-        assertEquals("Keycloak SSO", service.getName());
-        assertEquals("keycloak", service.getIdentifier());
+        //when
+        descriptionFactory.applyData(file, null, landscapeDescription);
+
+        //then
+        Map<String, GroupDescription> groups = landscapeDescription.getGroups();
+        assertThat(groups).isNotEmpty();
+
+        GroupDescription ingress = groups.get("ingress");
```
bonndan, line 109: @mrgamedev07 here we test for a new group definition that comes from a referenced file

src/main/java/de/bonndan/nivio/input/SourceReferencesResolver.java
```diff
@@ -47,7 +47,7 @@ public void resolve(final LandscapeDescription landscapeDescription) {
             }
 
             try {
-                landscapeDescription.addItems(factory.getDescriptions(ref, baseUrl));
+                factory.applyData(ref, baseUrl, landscapeDescription);
```
mrgamedev07, line 50: variable name factory might be changed. It confuses with the spring bean factory.
 
``InputFormatHandler  inputFormatHandlerFactory;                                                              
inputFormatHandlerFactory.applyData(ref, baseUrl, landscapeDescription);``
This should be read able .i guess !!!

src/test/java/de/bonndan/nivio/input/KubernetesTest.java
```diff
@@ -52,20 +53,26 @@ void setup() {
     }
 
     @Test
-    public void testRead() throws IOException {
+    public void testRead() {
```
mrgamedev07, line 56: Just added a comment here, for confirmation .you have removed IOException here. are you sure?

src/test/java/de/bonndan/nivio/input/dto/InputFormatHandlerNivioTest.java
```diff
@@ -16,29 +16,35 @@
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.*;
```
mrgamedev07, line 22: Import * again ;

Issue #449, opened 2021-02-20
User manual cleanup


bonndan:
- [x] Need to improve the response when there is no manual available.
- [x] It will look better if the "This manual Page doesn't exist " message appears  in the middle, centre aligned and  top of the page 
- [x] And why we need a scroller if the content is just a single line? May be logic should be dynamically shown scroll bar when the page does have a manual content 
![image](https://user-images.githubusercontent.com/73687196/108592049-ed159980-7391-11eb-909e-29657d4de63f.png)

_Originally posted by @mrgamedev07 in https://github.com/dedica-team/nivio/pull/428#discussion_r579635276_

PR #455, opened 2021-02-24
449 Manual Message style changed

mrgamedev07:
#449 scroll bar appears dynamically and Message aligned center


bonndan (changes requested):
Please add a simple test.

PR #465, opened 2021-03-04
430 forward to landscape

### LINKED ISSUE #430: Forward user to landscape if only single one exists
There is no point showing the front page with the landscape selection if only one landscape is configured. Instead, forward the user to the landscape.

mrgamedev07:
closes #430 
Now when there is only one landscape present, ui will be routed to the landscape map page .

### DIFF
--- a/src/main/app/src/Components/Landscape/Overview/Overview.tsx
+++ b/src/main/app/src/Components/Landscape/Overview/Overview.tsx
@@ -18,11 +18,14 @@ const Overview: React.FC<Props> = ({ setSidebarContent, setPageTitle }) => {
   const [landscapes, setLandscapes] = useState<ILandscape[]>([]);
   const [landscapeLinks, setLandscapeLinks] = useState<ILandscapeLinks | null>();
   const [loadLandscapes, setLoadLandscapes] = useState<boolean>(true);
+  const [landscapesCount, setLandscapesCount] = useState<Number | null>();
+
 
   const getLandscapes = useCallback(async () => {
     if (loadLandscapes) {
       setLandscapeLinks(await get('/api/'));
       if (landscapeLinks) {
+        let count = 0;
         for (const landscapeLink in landscapeLinks._links) {
           const landscapeDescription: ILandscape | null = await get(
             landscapeLinks._links[landscapeLink].href
@@ -30,7 +33,9 @@ const Overview: React.FC<Props> = ({ setSidebarContent, setPageTitle }) => {
           if (landscapeDescription) {
             setLandscapes((oldLandscapes) => [...oldLandscapes, landscapeDescription]);
           }
+          count++;
         }
+        setLandscapesCount(count);
       }
       setLoadLandscapes(false);
     }
@@ -42,7 +47,7 @@ const Overview: React.FC<Props> = ({ setSidebarContent, setPageTitle }) => {
     setPageTitle('All Landscapes');
   }, [getLandscapes, setSidebarContent, setPageTitle]);
 
-  return <OverviewLayout landscapes={landscapes} setSidebarContent={setSidebarContent} />;
+  return <OverviewLayout landscapes={landscapes} setSidebarContent={setSidebarContent} landscapesCount={landscapesCount} />;
 };
 
 export default Overview;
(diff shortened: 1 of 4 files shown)


src/main/app/src/Components/Landscape/Overview/OverviewLayout.tsx
```diff
@@ -40,86 +40,93 @@ const useStyles = makeStyles((theme: Theme) =>
...
 
 /**
  * Displays all available landscapes and provides all needed navigation
  */
 
-const OverviewLayout: React.FC<Props> = ({ landscapes, setSidebarContent }) => {
+const OverviewLayout: React.FC<Props> = ({ landscapes, setSidebarContent, landscapesCount }) => {
   const classes = useStyles();
   const componentClasses = componentStyles();
+  const history = useHistory();
```
bonndan, line 53: Is it possible to use history in the parent component? So we would not have to pass the count to Overviewlayout.

src/main/app/src/Components/Landscape/Overview/OverviewLayout.test.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
```
bonndan, line 1: 👍

src/main/app/src/Components/Landscape/Overview/OverviewLayout.tsx
```diff
@@ -40,86 +40,93 @@ const useStyles = makeStyles((theme: Theme) =>
...
-    content = landscapes.map((landscape) => {
-      let itemCount = 0;
-      landscape.groups?.forEach((group) => (itemCount += group.items.length));
-      let stats =
-        itemCount +
-        ' items in ' +
-        (landscape.groups ? Object.keys(landscape.groups).length : 0) +
-        ' groups';
+  if (Array.isArray(landscapes) && landscapes.length && landscapesCount) {
+    if (landscapesCount > 1) {
```
bonndan, line 57: If needed at all, I would revert the if statement so that the "else" is not necessary:

if (count ==1) {
history....
return null;
}

src/main/app/src/Components/Landscape/Overview/Overview.tsx
```diff
@@ -18,19 +18,24 @@ const Overview: React.FC<Props> = ({ setSidebarContent, setPageTitle }) => {
...
   const [landscapeLinks, setLandscapeLinks] = useState<ILandscapeLinks | null>();
   const [loadLandscapes, setLoadLandscapes] = useState<boolean>(true);
+  const [landscapesCount, setLandscapesCount] = useState<Number | null>();
+
 
   const getLandscapes = useCallback(async () => {
     if (loadLandscapes) {
       setLandscapeLinks(await get('/api/'));
       if (landscapeLinks) {
+        let count = 0;
```
bonndan, line 28: Is the count necessary at all? Why not use size/length?

bonndan (changes requested):
Please see my remarks

src/main/app/src/Components/Landscape/Overview/Overview.tsx
mrgamedev07: yes, the count is necessary because landscapeLinks._links -data from backend is JSON object instead of JSON array, so I can't use size or length. yet afraid to change from backend as it might affect the main functionality and requires big change.

src/main/app/src/Components/Landscape/Overview/OverviewLayout.tsx
mrgamedev07: yes, I tried this but the problem here is the call to OveriewLayout is synchronous, which means whenever a single landscape is loaded OveriewLayout is getting called .we are not waiting to fetch all landscape and to display at once. so it is always satisfying if condition and "history.push" happens every time.

src/main/app/src/Components/Landscape/Overview/Overview.tsx
bonndan: `Object.keys(obj).length` would this work? https://stackoverflow.com/questions/126100/how-to-efficiently-count-the-number-of-keys-properties-of-an-object-in-javascrip

bonndan (approved):
nice!

Issue #438, opened 2021-02-15
Item focus does not center item


bonndan:
Depending on map size or item position, the scroll distance when clicking on the focus button differs from the required distance, so that often items are not centered properly.

relates to #471

PR #469, opened 2021-03-09
438 disable centering

bonndan:


mfbieber (approved):
Looks fine for a quick disabling. If we don't fix it, we should remove the commented code then. Do we have a ticket for a fix?

bonndan:
#438 is kept open until the behavior is fixed. For interactive use of nivio I'd really like to ahve it. Especially in large maps.

PR #478, opened 2021-03-13
436 notification context

bonndan:
* After successful processing the landscape map is reloaded.
* The latest event is show in the bottom status bar.
* "Show map" button has been removed.

towards #436


src/main/java/de/bonndan/nivio/notification/MessagingService.java
```diff
@@ -29,11 +31,25 @@ public MessagingService(SimpMessagingTemplate template) {
...
+    @EventListener(ProcessingFinishedEvent.class)
+    public void onProcessingEvent(ProcessingEvent processingEvent) {
         EventNotification eventNotification = EventNotification.from(processingEvent);
         fifo.add(eventNotification);
-        LOGGER.info("Broadcasting processing event: " + processingEvent.getType());
+        LOGGER.debug("Broadcasting processing event: " + processingEvent.getType());
+        this.template.convertAndSend(WebSocketConfig.TOPIC + EVENTS, eventNotification);
+    }
+
+    @EventListener(InputChangedEvent.class)
```
mfbieber, line 42: I couldn't find who triggered this event, then I saw the changes in `LocalFileObserver`, but these are not in this branch. I guess the commit to develop was not on purpose?

Anyhow, didn't we send an event via websockets on `InputChangedEvent` before? Only on `ProcessingFinishedEvent`? Does this change anything in the notification of the frontend?

src/main/app/src/Components/Notification/Notification.tsx
```diff
@@ -28,19 +25,8 @@ const Notification: React.FC = () => {
...
-                : `${notificationMessage.message || 'Event Error: No message received'}`,
-              key: new Date().getTime(),
-              landscape: notificationMessage.landscape,
-              level: notificationMessage.level,
-            };
-
-            setSnackPack((prevArray) => [...prevArray, snackPackMessage]);
-            setOpen(true);
+          if (notificationMessage.type === 'ProcessingFinishedEvent') {
+            notificationContext.next(notificationMessage);
```
mfbieber, line 29: This looks so much cleaner now 👍 👍

src/main/java/de/bonndan/nivio/notification/MessagingService.java
mfbieber: As I have seen so far, we only use `ProcessingFinishedEvent` in the frontend. What is `InputChangedEvent` for, especially sending it via websockets to the frontend?

mfbieber (approved):
The results of what happens in this PR are great. Well done 👍 😻

src/main/java/de/bonndan/nivio/notification/MessagingService.java
bonndan: InputChangedEvent is fired when files change, "long" before the processing has finished. It would notify the user that something will happen behind the curtains now. However, I will recheck that behavior again.

PR #479, opened 2021-03-13
441 consistent icons

### LINKED ISSUE #441: Consistent buttons for item/group selection and zooming
After having used the search the results appear with an item icon and a focus icon. The icon icon has no function, but could be used as focus instead.

bonndan:
Consisten clickable icons

image

closes #441


mfbieber (approved):
This looks fine 😄

Is it now possible to add your own colors for specific items? Or how does the "hex color code to override the inherited group color" express itself?

bonndan:
Exactly. Is was already possible before, but never documented :)

PR #482, opened 2021-03-16
436 processing changes

bonndan:
This PR allows tracking of landscape component (Group, Item, Relation) changes on field level in order to highlight changes in the UI.

requirement for #436


src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
```diff
@@ -18,69 +18,129 @@ protected ItemRelationProcessor(ProcessLog processLog) {
...
-                    } else {
-                        relation.getSource().getRelations().remove(relation);
-                        relation.getSource().getRelations().add(relation);
+
+                if (!isValid(relationDescription, landscape)) {
+                    continue;
+                }
+
+                Optional<Relation> existing = getExisting(relationDescription, landscape, origin);
+                if (existing.isPresent()) {
```
mfbieber, line 37: I think we can find better names for `existing` and `getExisting()` 😸 - `if(existing.isPresent())` should be present by definition 😄 

How about just `getRelation()`? If it is present, it is existent.

mfbieber: I understand why you chose `existing`, below we have `created`, but still it reads strangely like it is now :)

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
```diff
@@ -18,69 +18,129 @@ protected ItemRelationProcessor(ProcessLog processLog) {
...
+                    continue;
+                }
+
+                Relation created = create(relationDescription, landscape);
+                affected.add(created);
+                processed.add(created);
+                processLog.info(String.format(origin + ": Adding relation between %s and %s", created.getSource(), created.getTarget()));
+                changelog.addEntry(created, ProcessingChangelog.ChangeType.CREATED, null);
             }
+
```
mfbieber, line 55: Also, what you want do do here is use the Optionals a little better, e.g. refactor to:
````
                Optional<Relation> optionalRelation = getRelation(relationDescription, landscape, origin);
                optionalRelation.ifPresentOrElse((relation) -> {
                        Relation update = update(relationDescription, relation);
                        affected.add(update);
                        processed.add(update);
                        processLog.info(String.format("Updating relation between %s and %s", update.getSource(), update.getTarget()));
                        List<String> changes = relation.getChanges(update);
                        if (!changes.isEmpty()) {
                            changelog.addEntry(update, ProcessingChangelog.ChangeType.UPDATED, String.join(";", changes));
                        }
                    }, 
                        () -> { 
                        Relation created = create(relationDescription, landscape);
                        affected.add(created);
                        processed.add(created);
                        processLog.info(String.format(origin + ": Adding relation between %s and %s", created.getSource(), created.getTarget()));
                        changelog.addEntry(created, ProcessingChangelog.ChangeType.CREATED, null);
               });
````
And then you could extract two methods and obtain:
````
                Optional<Relation> optionalRelation = getRelation(relationDescription, landscape, origin);
                optionalRelation.ifPresentOrElse(
                        (relation) -> updateRelation(
                                changelog, processed, affected, relationDescription, relation
                        ), 
                        () -> createRelation(
                                landscape, changelog, processed, origin, affected, relationDescription
                        )
                );
````

File: src/main/java/de/bonndan/nivio/input/ProcessingChangelog.java [hunk 3]
```diff
@@ -0,0 +1,146 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+@Schema(description = "Changelog for a single landscape processing run (update of landscape from different sources).")
+public class ProcessingChangelog {
+
+    @Schema(description = "The key FullyQualifiedIdentifier of a component")
+    final Map<String, Entry> changes = new HashMap<>();
```
mfbieber, line 18: I was wondering, who is reading from the ProcessingChangelog? What is the purpose of the object?

src/main/java/de/bonndan/nivio/model/ComponentDiff.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param b       second
+     * @param key     label
+     * @param changes array of changes
+     */
+    @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+    public static void compareOptionals(Optional<Object> a, Optional<Object> b, String key, List<String> changes) {
+        if (a.isEmpty() && b.isEmpty()) {
+            return;
+        }
+        if (a.isPresent() && b.isEmpty() || a.isEmpty() && b.isPresent()) {
```
mfbieber, line 37: `b.isPresent()` is always true when reached, because it has been checked by the preceding conditions. Still, I would leave it there for better reliability of the code. I would suggest suppressing the warning and add a small comment to it.

src/main/java/de/bonndan/nivio/model/ComponentDiff.java
```diff
@@ -0,0 +1,61 @@
...
+            changes.add(key + " changed to: " + b.orElse(""));
+            return;
+        }
+        if (!org.apache.commons.lang3.StringUtils.equals(a.map(Object::toString).orElse(""), b.map(Object::toString).orElse(""))) {
```
mfbieber, line 41: Here we use the Apache Commons `StringUtils`, below we use the Spring `StringUtils` for `.isEmpty()` (line 57). Why don't we stick to just Apache Commons?

src/main/java/de/bonndan/nivio/model/ComponentDiff.java
```diff
@@ -0,0 +1,61 @@
...
+     * Compares changes in two strings.
+     *
+     * @param a       first
+     * @param b       second
+     * @param key     label
+     * @param changes array of changes
+     */
+    public static void compareStrings(String a, String b, String key, List<String> changes) {
+        if (!org.apache.commons.lang3.StringUtils.equals(a, b)) {
+            changes.add(key + " changed to: " + b);
```
mfbieber, line 20: Any thoughts on thread safety and reads and writes to the changes `List` in nivio? Does it matter and could we run into problems?

mfbieber (approved):
Looks fine overall 👍

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: changed to "current"

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: nice 👍

src/main/java/de/bonndan/nivio/model/ComponentDiff.java
bonndan: I prefer Spring's stuff, but here I was lookingfor a null safe equality check. It turns out however that this is not necessary. Refactored.

src/main/java/de/bonndan/nivio/model/ComponentDiff.java
bonndan: That's why I left it there. Refactored to clearify a bit

src/main/java/de/bonndan/nivio/model/ComponentDiff.java
bonndan: Maybe. Ad modifying changes is a side effect. Since every compare method returns void, I'll rewrite them to return changes instead.

src/main/java/de/bonndan/nivio/input/ProcessingChangelog.java
bonndan: It was the first attempt to provide a log to users so that they can track changes.

PR #488, opened 2021-03-19
Improve search

### LINKED ISSUE #302: Improve search UX

bonndan:
closes #302 

* various improvements of search term handling
* improved facets


src/main/app/src/Components/Landscape/Search/Search.tsx
```diff
@@ -54,56 +64,122 @@ const Search: React.FC<PropsInterface> = ({ setSidebarContent, ...props }) => {
...
 
+    const help = (
+      <div>
+        <h3>Search</h3>
+        Start typing to match items having name parts or description starting with the given term.
+        Find everything starting with 'foo':
+        <br />
+        <pre>foo</pre>
+        <br />
+        Use whitespaces as AND condition (default). Search in name and description for both words:
```
mfbieber, line 131: What would be the description? 

I tried to select several Wordpress items by typing `wor c` in http://localhost:3000/#/landscape/nivio:example and I was expecting to select wordpress-web and wordpress-redis and wordpress-mariadb, but it only selects wordpress-redis. They belong to the groups"content" and "common".
`wor co` does not find anything, for example. So the first word is the name of the item and the second is?

mfbieber (approved):
This improvement is great. I also like that you can select e.g. component, layer, etc. to find things 👍 

Just one question about the tooltip for the search (see in the comments), maybe we could make that clearer to understand.

src/main/app/src/Components/Landscape/Search/Search.tsx
bonndan: Good point. The "multi field" query regards identifier, name and description. If you want to search in groups, you have to enter "group:co*". In addition, you query translates to "wor* AND c* in the three fields.

That reminds me to add facets for item.type and item.group.

bonndan: I've added #498 as follow-up

bonndan: + #500 as follow-up

PR #497, opened 2021-03-27
439 removed items

### LINKED ISSUE #439: Removed items still visible after update cycle
It seems that items which have been removed from the input sources are not properly deleted in the landscape.

- [x] ensure deleted items do not appear anymore

This is a regression bug, maybe introduced by immutable components.

bonndan:
closes #439


mfbieber (changes requested):
The code looks good, but I still have to manually reload the frontend to see the changes that were made. 

Sometimes the frontend reloads after the next change has been made, but does not display the latest one. I think something is still off.

mfbieber:
https://github.com/dedica-team/nivio/pull/482 has the same change/reloading issue btw. The backend detects changes, but showing the results in the frontend lag behind by one change

mfbieber (approved):
Sorry, my mistake. Needed to rebuild the frontend with yarn install.

PR #505, opened 2021-04-02
436 showing change log

### LINKED ISSUE #436: Highlight changes after input data update

### LINKED ISSUE #504: Deleted relation is not reestablished
Once a relation has been deleted during a landscape change, it does not reappear in the map when reestablished.

bonndan:
closes #436 
fixes #504 

image


mfbieber (commented):
Sorry that it took me so long to review this.

I went rather briefly through the code since all of it looks very much okay 👍 

I just noticed three things:
- The icon for the changelog is the RSS feed symbol. I guess this is intentional. It reminded me a bit of a wifi connection symbol. Maybe we can find a better one? I'll try to make a suggestion.
- One has to click on the icon again to see a new change. So if I have the changelog opened and I make a new change, it does not reload with the new change.
- I cannot remove the changelog window again. But I guess this applies to everything in the panel on the right side?

PR #517, opened 2021-04-16
[#516] fix that default group "common" has no landscape identifier

### LINKED ISSUE #516: Group "common" is not clickable in map (lacking id)
When navigating the demo landscape "example" the group "common" cannot be clicked because attributes are missing in the SVG.

image

bonndan:
closes #516


DanielBruenker (approved):
Very good! I tested it locally and it works fine again.

PR #522, opened 2021-04-22
343 dot input

### LINKED ISSUE #343: GraphViz/Dot Language as source input format
The dot language describes graphs and is widely use to automatically create simpler graph drawing (and complex one, like plantuml does). Nivio could parse dot files and translate nodes and edges into items and relations.

See https://github.com/nidi3/graphviz-java#user-content-parsing (this library could be used as-is).

- [x] check if graphviz nodes can have arbitrary attributes which could be translated into item fields/labels.

### LINKED ISSUE #524: Item labels are not updated on input change
When an ItemDescription DTO is used to create an updated copy of an existing landscape item, the labels are not updated at all.

It is expected that the labels are replaced.

bonndan:
closes #343 
fixes #524

### DIFF
--- a/docs/source/input.rst
+++ b/docs/source/input.rst
@@ -82,7 +82,12 @@ enhance items and relations. However, it is necessary to prefix attributes that 
    :linenos:
 
     digraph G {
-            main [nivio_owner = Marketing, nivio_software="Wordpress 2.0", nivio_group=FooBar]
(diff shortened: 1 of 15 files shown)


bonndan:
@mfbieber could you start with this review please because it contains a bugfix. Thx!

PR #566, opened 2021-06-09
[#231] ArchUnit 

robrobert99:
No beans  in Application class

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,13 @@
 
         <dependency>
             <groupId>com.tngtech.archunit</groupId>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -98,6 +98,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
bonndan, line 101: https://github.com/societe-generale/arch-unit-maven-plugin#how-to-use-archunit-maven-plugin-

You still need to configure some tests.

bonndan (changes requested):

bonndan (approved):

PR #580, opened 2021-06-20
[#574] replaced deprecated Lucene RamDirectory

### LINKED ISSUE #574: IndexNotFoundException in Apache Lucene lib
**Describe the bug**
IndexNotFoundException occurs during program execution

**To Reproduce**
1. Execute program with DEMO=1
2. Open application in a web browser 

**Additional context**
It seems the exception appears random, during program use.

[Exception.txt](https://github.com/dedica-team/nivio/files/6662466/Exception.txt)

bonndan:
hopefully closes #574 

If the error persists we need to make SearchIndex instances persisting (e.g. using a repo). The fix is to make search index location predictable using the landscape identifier. Beforehands a new (random?) location was used with every new instance.


robrobert99 (approved):
The replacement of Lucene RamDirectory is working. I had no Exceptions while executing nivio.

PR #609, opened 2021-07-09
[#564] framework key to freetext search

### LINKED ISSUE #564: Add frameworks to free text search
As a user I want to enter "java" in the search in order to find items having java as framework. 

At the moment java in only searchable using the "framework" facet.

- [x] add all framework keys (not values) to full text search
- [x] ensure a framework can be found

bonndan:
closes #564


src/main/java/de/bonndan/nivio/search/SearchDocumentFactory.java
```diff
@@ -64,7 +71,7 @@ public static Document from(@NonNull final Item item, @Nullable List<StatusValue
         }
 
         Document document = new Document();
-
+        List<String> genericStrings = new ArrayList<>();
```
robrobert99, line 74: Move this variable down, where it is used.

robrobert99 (approved):

PR #643, opened 2021-08-12
relation labels using a "label" label

bonndan:
image

* Displays labels on relations.
* Updates for the "internals" demo landscape


src/main/java/de/bonndan/nivio/output/map/svg/SVGRelation.java
```diff
@@ -133,14 +132,13 @@ private ContainerTag addAttributes(ContainerTag g, Relation relation) {
         return g;
     }
 
-    private ContainerTag label(BezierPath bezierPath, String fillId) {
+    private ContainerTag label(String text, BezierPath bezierPath, String fillId) {
```
robrobert99, line 135: Add @Nullable or return Optional.

bonndan: changed...

robrobert99 (approved):

Issue #275, opened 2020-09-13
Add kubernetes change observer


bonndan:
add an observer implementation for kubernetes

PR #646, opened 2021-08-13
Event observer

robrobert99:
New feature related to #275. 
This feature needs tests and further improvement, before being merged.


src/main/java/de/bonndan/nivio/input/Seed.java
```diff
@@ -67,12 +68,13 @@ private URL asURL(String s) {
...
             return demoFiles;
         }
+        String value = ConfigurableEnvVars.DEMO.value().orElse("");
 
         Path currentRelativePath = Paths.get("");
         String absPath = currentRelativePath.toAbsolutePath().toString();
         try {
-            demoFiles.add(new File(absPath + "/src/test/resources/example/pet_clinic.yml").toURI().toURL());
-            if (demo.equalsIgnoreCase("all")) {
+            demoFiles.add(new File(absPath + "/src/test/resources/example/example_k8s.yml").toURI().toURL());
```
bonndan, line 76: Nope 😄 

We should introduce that the demo file can be chosen by name instead of "all" or "1"

src/main/java/de/bonndan/nivio/input/kubernetes/KubernetesObserver.java
```diff
@@ -0,0 +1,77 @@
...
+public class KubernetesObserver implements InputFormatObserver {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KubernetesObserver.class);
+
+    private final Landscape landscape;
+    private final ApplicationEventPublisher eventPublisher;
+    private final KubernetesClient kubernetesClient;
+    private final List<Long> eventUidList;
+
+    public KubernetesObserver(@NonNull final Landscape landscape,
```
bonndan, line 29: Can we guarantee or ensure that this is NotNull?

src/main/java/de/bonndan/nivio/input/kubernetes/KubernetesObserver.java
```diff
@@ -0,0 +1,77 @@
...
+                change = true;
+                LOGGER.info("K8s change detected");
+                eventPublisher.publishEvent(new InputChangedEvent(new ObservedChange(landscape, "k8s cluster changed")));
+            }
+        }
+
+    }
+
+    @NonNull
+    private List<Long> getK8sComponents() {
```
bonndan, line 53: If we keep a list or map of the original items we could detect where the change happened.

src/main/java/de/bonndan/nivio/input/kubernetes/KubernetesObserver.java
```diff
@@ -0,0 +1,77 @@
...
+        this.kubernetesClient = kubernetesClient;
+        this.eventPublisher = eventPublisher;
+        this.eventUidList = getK8sComponents();
+    }
+
+    @Override
+    public void run() {
+        var change = false;
+        while (!change) {
+            sleep(1000);
```
bonndan, line 42: The observers are started in a thread pool with scheduleWithFixedDelay. This basically means that we do not have to sleep here. However, the given delay (30s I think) might be a bit long. Feel free to change the delay (maybe on a per-observer basis).

Issue #647, opened 2021-08-13
Failure when removing kubernetes items


robrobert99:
**Describe the bug**
RuntimeException during the removing of items.

**To Reproduce**
1. Checkout the event_observer branch
2. Scale down a deployment two times

**Expected behavior**
Elements should be removed without error.

**Screenshots**
![image](https://user-images.githubusercontent.com/84507051/129369219-0fa96e77-7bb5-4049-8e1f-fb661fcf81ab.png)

PR #651, opened 2021-08-21
[#647] handling missing relation ends gracefully

bonndan:


src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
```diff
@@ -45,7 +49,7 @@ public ProcessingChangelog process(@NonNull final LandscapeDescription input, @N
                         })
                         .orElseGet(() -> {
                             Relation created = RelationFactory.create(origin, relationDescription, landscape);
-                            processLog.info(String.format(origin + ": Adding relation between %s and %s", created.getSource(), created.getTarget()));
+                            processLog.info(String.format("%s: Adding relation between %s and %s", origin, created.getSource(), created.getTarget()));
```
dudlsack, line 52: @bonndan wenn hier Performance wichtig ist, würde ich die Strings konkatenieren, auch wenn dadurch die Lesbarkeit leidet.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
```diff
@@ -66,29 +70,39 @@ public ProcessingChangelog process(@NonNull final LandscapeDescription input, @N
...
+     *
+     * @param landscape the current landscape
+     * @param relation the relation to remove
+     * @param relationEnd the relation source or target
+     */
+    private void removeRelationFromItem(Landscape landscape, Relation relation, Item relationEnd) {
+        var fqi = relationEnd.getFullyQualifiedIdentifier();
+        landscape.getItems().find(fqi.getItem(), fqi.getGroup())
+                .ifPresentOrElse(item -> {
+                    if (!item.removeRelation(relation)) {
```
dudlsack, line 94: Um die Lesbarkeit zu verbessern würde ich item.removeRelation(relation) in eine sprechende Variable auslagern und die dann in der If-Bedingung verwenden

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
```diff
@@ -66,29 +70,39 @@ public ProcessingChangelog process(@NonNull final LandscapeDescription input, @N
...
 
+    /**
+     * Gracefully finds the relation end item in the landscape and tries to remove the relation.
+     *
+     * @param landscape the current landscape
+     * @param relation the relation to remove
+     * @param relationEnd the relation source or target
+     */
+    private void removeRelationFromItem(Landscape landscape, Relation relation, Item relationEnd) {
+        var fqi = relationEnd.getFullyQualifiedIdentifier();
```
dudlsack, line 91: Wie wäre es mit identifier anstatt fqi? Bin ein Freund von sprechenden Variablen.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
```diff
@@ -66,29 +70,39 @@ public ProcessingChangelog process(@NonNull final LandscapeDescription input, @N
...
     }
 
+    /**
+     * Gracefully finds the relation end item in the landscape and tries to remove the relation.
+     *
+     * @param landscape the current landscape
+     * @param relation the relation to remove
+     * @param relationEnd the relation source or target
+     */
+    private void removeRelationFromItem(Landscape landscape, Relation relation, Item relationEnd) {
```
dudlsack, line 90: Wie wäre es wenn du direkt die Items injectest anstatt Landscape, denn ich habe mich zuerst gefragt warum Landscape gebraucht wird. Dann könnte die Methode auch removeRelationFromItems() heißen.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: Good point, but here we don't use string concatenation very often. We'll keep it in mind when tackling #688.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: Item already has a field `identifier` which is part of the fully qualified identifier (but not unique). So that might be misleading.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: Actually we pick one item only here, the whole collection is never accessed

dudlsack: Sure it is in line 92 `landscape.getItems()`. 

If you have a signature like `private void removeRelationFromItem(List<Item> items, Relation relation, Item relationEnd)` the method call `removeRelationFromItem(landscape.getItems(), relation, relation.getSource());`would be more readable.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
dudlsack: This is not my point here ;-) Here you call a method within an If clause which executes business logic and not only does a check. In my opinion this is bad practice and not really readable. You could use a variable like `isRelationRemoved` for example and use it in the If clause.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: `landscape.getItems()` returns an object, not a collection. We use an in-memory (CQEngine) here. I will consider rewriting the method.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: Item already has a field identifier which is part of the fully qualified identifier (but not unique). So that might be misleading.

src/main/java/de/bonndan/nivio/input/ItemRelationProcessor.java
bonndan: Sorry, above I have answered the question below 🤦

Issue #496, opened 2021-03-26
Show build number 


bonndan:
As a user I need to see the build number to report issues properly.

- [] read git commit id: https://doc.akka.io/docs/akka-management/current/kubernetes-deployment/building-using-maven.html#building-using-maven
- [x] show in API root: store version and commit id in application.yml -> https://www.baeldung.com/spring-boot-build-properties#yaml
- [x] show in frontend
- [ ] ~show in manual if possible (see if index.rst can be modified with placeholders like application.yml)~

bonndan:
Using a dynamic build number with Sphinx/rst is not possible in the scope of this ticket. It would require too much effort if possible at all.

PR #689, opened 2021-09-08
#496 maven git commit id plugin

jenarp:
Please have a look @bonndan


pom.xml
```diff
@@ -533,6 +557,8 @@
                     </dependency>
                 </dependencies>
             </plugin>
+
```
bonndan, line 560: please remove these line feeds

src/main/resources/application.yml
```diff
@@ -55,6 +56,8 @@ nivio:
   demo: ${DEMO:}
 
 
+
```
bonndan, line 59: please remove

src/main/resources/application.yml
```diff
@@ -86,4 +89,6 @@ spring:
   profiles: local
 
 server:
-  port: 8080
\ No newline at end of file
+  port: 8080
+
```
bonndan, line 93: please remove one lf

bonndan (changes requested):
Looks good, but please remove the newly introduced line feeds where not necessary.

bonndan (approved):
👍

PR #710, opened 2021-10-06
705 one wording

robrobert99:


bonndan:
A simple test for the expected date string might be useful.

bonndan (changes requested):

jenarp (approved):

Issue #717, opened 2021-10-15
Fix landscape logo in SVG


bonndan:
At the moment the landscape logo cannot be shown in the SVG because it gets lost during processing.

Acceptance criteria

- [x] icon is visible in SVG (e.g. print view)

Fixes

- [x] fix that the "icon" is correctly copied from dto to the landscape models and from old to new landscape models (requires new constructor param)
- [x] fix that AppearanceProcessor correctly sets the `_iconData `label (resolved icon url) to the landscape
- [x] fix that SVGDocument reads _iconData instead of using `getIcon()`

PR #747, opened 2021-10-29
717 fix landscape logo

jenarp:
Dear Team,

please have a look. Ticket #717 .

Happy weekend!


src/main/java/de/bonndan/nivio/input/dto/LandscapeDescription.java
```diff
@@ -291,11 +288,13 @@ public void setLabels(Map<String, String> labels) {
     }
 
     @Override
+    @JsonAnyGetter
     public String getLabel(String key) {
```
robrobert99, line 292: @JsonAnyGetter shouldn't be here. According to the documentation a Map must be returned.
https://fasterxml.github.io/jackson-annotations/javadoc/2.6/com/fasterxml/jackson/annotation/JsonAnyGetter.html

bonndan: 👍 my bad.

Please move it to the `getLabels()` method.

robrobert99 (approved):

bonndan (approved):

PR #748, opened 2021-11-02
Add null filter

robrobert99:


bonndan:
@robrobert99 please also test this with "icon", which is where the error originally appeared.

bonndan (approved):

jenarp (approved):

PR #749, opened 2021-11-03
Frontend mapping

robrobert99:


src/main/app/src/App.tsx
```diff
@@ -1,139 +1,145 @@
...
 
 interface Config {
-  baseUrl: string;
-  version: string;
-  brandingForeground: string;
-  brandingBackground: string;
-  brandingSecondary: string;
-  brandingLogoUrl: string;
-  brandingMessage: string;
+    baseUrl: string;
```
bonndan, line 16: @robrobert99 @jenarp  Is your IDE configured to regard the `.prettierrc` config file?

src/main/app/package.json
```diff
@@ -24,6 +24,7 @@
     "fs-extra": "^9.1.0",
     "html-react-parser": "^0.14.3",
     "react": "^16.14.0",
+    "react-async": "^10.0.1",
```
bonndan, line 27: Do we need that now?

src/main/app/src/Context/FrontendMappingContext.tsx
```diff
@@ -0,0 +1,26 @@
+import React, {useState} from 'react';
```
bonndan, line 1: This needs a test, too.

src/main/app/src/Context/FrontendMappingContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const FrontendMappingContext = React.createContext<FrontendMappingContextType>({
+    frontendMapping: new Map<string, string>()
+});
+
+const FrontendMappingProvider: React.FC = ({children}) => {
+    const [frontendMapping, setFrontendMapping] = useState(new Map);
+
+    get(`/api/mapping`).then((response) => {
```
bonndan, line 16: According to @DanielBruenker we should load that in a useEffect hook.

Issue #704, opened 2021-10-05
"Networks" missing at "INFO" in GUI


jenarp:
Referring to first task ticket #611.

"Networks" missing at "INFO" in GUI (e.g. @ "Blog")
It appears in report and in search bar of GUI but not in INFO of items.

Acceptance criteria:
- [x] consider if networks should be visible at "INFO"
- [x] if yes: add networks to it
- [x] stick to one wording: i.e. networks or network (I would like to discuss this with team as all facet names are singular)

![image](https://user-images.githubusercontent.com/67540243/135985606-d3a3c8ce-d719-476d-a14c-8f2494060d59.png)

![Screenshot 2021-10-05 at 10 14 32](https://user-images.githubusercontent.com/67540243/135985822-574e118c-a252-4d03-8d2e-bf06cc04aab1.png)

![image](https://user-images.githubusercontent.com/67540243/135985916-e49026ea-4658-40b1-ab42-66ed7a53168a.png)

PR #760, opened 2021-11-09
704 add networks in Gui

jenarp:
Ticket #704
last checkbox item not implemented yet, as I would like to discuss this with team first


src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
```diff
@@ -257,6 +258,13 @@ const Item: React.FC<Props> = ({ fullyQualifiedItemIdentifier, small }) => {
         ))
       : null;
 
+  // const networks =
```
bonndan, line 261: please remove

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
```diff
@@ -222,6 +222,7 @@ const Item: React.FC<Props> = ({ fullyQualifiedItemIdentifier, small }) => {
     : null;
   const assessmentStatus = assessments ? getItemAssessments(assessments) : [];
   const frameworks: ReactElement | null = item ? getLabelsWithPrefix('framework', item) : null;
+  const networks: ReactElement | null = item ? getLabelsWithPrefix('networks', item) : null;
```
bonndan, line 225: As discussed this method is not suitable, since keys and values are similar like in tags. This displays as:

image

src/main/app/src/labels.json
```diff
@@ -1,22 +1 @@
-{
```
bonndan, line 1: please revert if possible

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
```diff
@@ -413,6 +421,13 @@ const Item: React.FC<Props> = ({ fullyQualifiedItemIdentifier, small }) => {
                 </div>
               ) : null}
 
+              {networks ? (
+                <div className='networks'>
+                  <Typography variant={'h6'}>Networks</Typography>
```
bonndan, line 426: We should start to use frontend mapping where possible

bonndan (changes requested):

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
jenarp: Yep, I know. But then I thought it makes sense to have it as in frameworks while content is only a placeholder for maybe "network1":  "LAN", "network2: "VPN". But while writing this I realize that network is redundant. So, I will remove it as in tags... ;)

robrobert99 (approved):

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
jenarp: For me it was simpler to change the method getLabelsWithPrefix than to do it as with 'tags'.
I added a regexp to getLabelsWithPrefix looking if [key = prefix.value]: value --> only show primary 'content'
hope this is ok for you...

bonndan: Unfortunately no. The regex + 150 chars approach makes it brittle. Please see https://github.com/dedica-team/nivio/pull/760/commits/fbd70551b6a14e582a3c259d825ebff3e3ff32cc as a starting point.

jenarp: Added networks to ItemApiModel.

src/main/app/src/Components/Landscape/Modals/Item/Item.tsx
jenarp: Should frontend mapping only append to "networks" or also to "frameworks", "labels", etc. for consistency?
If yes, Robert and I would suggest to create a dedicated ticket.
To be discussed.

bonndan: 👍 a dedicated ticket sounds reasonable

src/main/java/de/bonndan/nivio/model/Item.java
```diff
@@ -321,5 +321,24 @@ public String getAssessmentIdentifier() {
         return changes;
     }
 
+    public String[] getNetworks() {
```
bonndan, line 324: We don't need this getter/setter, networks come via labels and ItemApiModel can use them. Labeled interface also has some convenience methods to pick prefixed labels.

bonndan (changes requested):

bonndan (approved):

Issue #656, opened 2021-08-24
Display group icon as hexagons


bonndan:
![image](https://user-images.githubusercontent.com/143695/130662920-4abaf5c8-e4e7-4ff0-91fa-87e13c5efd5d.png)

The marked icon should not appear as circle like a regular component. 

## Acceptance criteria:

The group icon should display 

- [ ] the configured group icon (if available=
- [ ] or as hexagons using ![image](https://user-images.githubusercontent.com/143695/141675450-3b8f9e91-9f95-4d3d-a184-0c53c76e6496.png)
- [ ] the icon to be used should be configured in the backend

PR #782, opened 2021-11-17
656 display group icon

jenarp:
Ticket #656


src/main/java/de/bonndan/nivio/output/icons/IconService.java
```diff
@@ -30,7 +28,7 @@ public IconService(LocalIcons localIcons, ExternalIcons externalIcons) {
 
     /**
      * Returns the proper icon url for an item.
-     *
+     * <p>
```
bonndan, line 31: Please ask @basseur how to remove this by configuring your IDE.

bonndan (approved):

src/main/java/de/bonndan/nivio/output/icons/IconService.java
jenarp: I figured it out myself. Credits go to Google... ;)

PR #883, opened 2022-04-09
Preparing security release 0.5.1

bonndan:
@jenarp please check out, build and test


jenarp (approved):
I could not find any functional problems, but some issues regarding style for which I write tickets for.

PR #914, opened 2022-05-17
Replace null values with empty string instead of discarding the change

MarvinSchoening:
Replace null values with empty string instead of discarding the changese and keeping the old values

- builder.withName() isn't set to empty string since its necessary for the landscape. Its only replaced with a new value

### DIFF
--- a/src/main/java/de/bonndan/nivio/model/LandscapeFactory.java
+++ b/src/main/java/de/bonndan/nivio/model/LandscapeFactory.java
@@ -88,17 +88,13 @@ public class LandscapeFactory {
         final boolean isPartial = input.isPartial();
 
         //overwrite some data which is not handled by resolvers
-        if (!isPartial || input.getContact() != null) {
-            builder.withContact(input.getContact());
-        }
-        if (!isPartial || StringUtils.hasLength(input.getName())) {
-            builder.withName(input.getName());
-        }
-        if (!isPartial || input.getDescription() != null) {
-            builder.withDescription(input.getDescription());
-        }
-        if (!isPartial || input.getOwner() != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/de/bonndan/nivio/util/SafeAssign.java
```diff
@@ -13,7 +14,7 @@ public class SafeAssign {
      * @param c consumer
      */
     public static void assignSafe(String s, Consumer<String> c) {
-        if (s != null) c.accept(s);
+        c.accept(Objects.requireNonNullElse(s, ""));
```
Matthimatiker, line 17: Doesn't this change the behavior?

Before, `accept()` was not called when `null` is provided.

MarvinSchoening: Yes we want to change the behaviour. It should save an empty string now instead of doing nothing if null is provided. Removing items from yaml completely had the side effect that the data didn't get removed in nivio. It was missleading.

jenarp (approved):

src/main/java/de/bonndan/nivio/model/LandscapeFactory.java
```diff
@@ -88,17 +88,13 @@ public static Landscape recreate(Landscape existing, LandscapeDescription input)
         final boolean isPartial = input.isPartial();
 
         //overwrite some data which is not handled by resolvers
-        if (!isPartial || input.getContact() != null) {
```
MarvinSchoening, line 91: hier morgen nochmal die logik anschauen, das umschreiben der logik war nicht korrekt so

